PR #22, opened 2026-05-27
Polish dictation HUD: hover, click pass-through, error states

andrewwashuta:
## Summary

Reworks the dictation HUD interaction layer end-to-end. The big shift is that stop-button hover detection and click pass-through both moved into a native Rust polling thread — WebKit throttles JS timers (including `rAF`) on the non-key HUD panel, which is what caused hover to only update on mouse-down.

- **Hover**: Rust thread polls cursor via `CGEvent` (no key-state dependency) and emits `hud-stop-hover` events; JS just listens and toggles `.is-hovered`. CSS `:hover` is intentionally removed to avoid sticky-hover after the panel resigns key.
- **Click pass-through**: same thread flips `set_ignore_cursor_events` based on whether the cursor is over the pill, so clicks in the transparent gutter around the HUD fall through to whatever app is underneath.
- **Stop button styling**: hover surface now uses `var(--accent)` (matches sidebar/icon buttons) instead of a bespoke white wash; default surface dropped from 10% → 5% foreground for a subtler resting state.
- **Padding / radii**: tightened HUD padding `4px → 3px` and bumped inner radii `5px → 6px` so they're perfectly concentric with the `10px` outer corner and close the optical gap on the right edge.
- **Error states**: both `silent-error` ("Nothing recorded") and `error` ("Error") now share the same text-only pill with a smoother 380ms decaying shake. Silent uses warm/burnt-orange tint, error uses destructive red. Both hold ~1.6s before the blur-fade exit.
- **Audio bars**: 5-bar layout, capped ambient bounce so room tone/typing still registers, faster snap-down on silence (carried over from earlier polish commit).

## Test plan

- [ ] Stop button hover updates while cursor moves over it (no click needed) and clears on cursor exit
- [ ] Clicking in the transparent area around the pill registers on the app behind the HUD (e.g. another text field stays focused)
- [ ] Clicking the stop button or drag handle still works normally
- [ ] No "auto-hover" highlight when the HUD reappears
- [ ] Dragging the HUD by the handle doesn't leave hover stuck
- [ ] Silent dictation triggers "Nothing recorded" pill with warm tint + shake
- [ ] Hard errors trigger "Error" pill with red tint + shake
- [ ] Both error pills persist ~1.6s before fading


andrewwashuta:
@greptile-apps

PR #27, opened 2026-05-28
Add theme preference setting

andrewwashuta:
## Summary
- New Appearance section in Settings with a System/Light/Dark segmented control, persisted to `localStorage` and applied via `data-theme` on `<html>` (init runs before render to avoid FOUC).
- Scopes the dark `@media (prefers-color-scheme: dark)` block to `:root:not([data-theme])` so an explicit choice wins over the OS preference in either direction.
- Nudges the microphone popover top offset by 2px so the selected item overlays the trigger label exactly — no visual jump on open.

## Test plan
- [ ] Toggle System / Light / Dark in Settings → Appearance and confirm the app switches immediately, persists across reload, and that Light wins over a dark OS pref (and vice versa).
- [ ] Open the microphone select and confirm the selected option's text doesn't shift relative to the trigger.

### DIFF
--- a/index.html
+++ b/index.html
@@ -4,6 +4,23 @@
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1.0" />
     <title>OS Notetaker</title>
+    <script>
+      // Pre-paint theme bootstrap. Keep in sync with src/lib/theme.ts; the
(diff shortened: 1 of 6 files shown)


andrewwashuta:
@greptile-apps

PR #47, opened 2026-05-29
build: deterministic scribe-api image (reproducible builds — Phase A)

jakubswierczek:
Phase A of the reproducible-builds plan (#45): make the `scribe-api` image bytes a deterministic function of the source, so anyone can rebuild a commit and get the same digest to check against the TEE attestation.

## Changes
- **Pin base images by digest** — `rust:1.95-bookworm@sha256:6258907…`, `debian:bookworm-slim@sha256:0104b33…` (resolved from Docker Hub; bump deliberately).
- **Strip absolute build paths** — `RUSTFLAGS=--remap-path-prefix=/app=. --remap-path-prefix=/usr/local/cargo=/cargo` + `CARGO_INCREMENTAL=0` in the builder stage.
- **CA bundle by copy, not apt** — `COPY --from=builder /etc/ssl/certs/ca-certificates.crt …` removes the only build-time network fetch and its version drift.
- **Commit-derived timestamps** — `SOURCE_DATE_EPOCH` (= `git log -1 %ct`) + buildkit `rewrite-timestamp=true`, and the `created` label now uses `git log -1 %cI` instead of `github.event.head_commit.timestamp` (which is push-only and non-deterministic on `workflow_dispatch`).

Already in the repo (not touched): `rust-toolchain.toml` (`1.95.0`), and `[profile.release]` with `codegen-units = 1` + `strip = true`.

## Risk / validation ⚠️
I can't run Docker/CI locally, so this **needs a CI build to validate**:
1. It still **compiles** (Dockerfile changes — esp. the `outputs:` form replacing `push:`/`tags:`, and the cert COPY path).
2. It actually **reproduces** — that's Phase B (a double-build + `diffoci` job), not proven here.

The `outputs: type=image,…,rewrite-timestamp=true` line is the most likely to need tweaking; it follows Docker's documented reproducible-build recipe. `build.outputs.digest` is still emitted (deploy is by tag, so it's informational).

## Not in scope
- Phase B (verify script + double-build CI guard) — follow-up.
- Digest-in-compose / prelaunch (anchor "c") — separate investigation.

### DIFF
--- a/.github/workflows/build-scribe-api.yml
+++ b/.github/workflows/build-scribe-api.yml
@@ -35,8 +35,11 @@ jobs:
       - name: Compute metadata
         id: meta
         run: |
-          sha_short=$(git rev-parse --short=7 HEAD)
-          echo "sha_short=${sha_short}" >> "$GITHUB_OUTPUT"
+          echo "sha_short=$(git rev-parse --short=7 HEAD)" >> "$GITHUB_OUTPUT"
+          # Reproducibility: derive timestamps from the commit, not build time,
+          # so a rebuild of the same commit yields the same image bytes.
+          echo "epoch=$(git log -1 --pretty=%ct)" >> "$GITHUB_OUTPUT"
+          echo "created=$(git log -1 --pretty=%cI)" >> "$GITHUB_OUTPUT"
 
       - uses: docker/setup-buildx-action@v3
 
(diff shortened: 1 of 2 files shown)


jakubswierczek:
Superseded by #45, which now consolidates the plan + Phase A + anchor (b) into one PR.

PR #48, opened 2026-05-29
ci: record deployed image digest as a git tag (reproducibility anchor b)

jakubswierczek:
Anchor **(b)** from the reproducible-builds plan (#45). We deploy by tag (dstack can't verify digest refs — see #46), so the TEE attestation binds a *tag*, not a content digest. This adds an in-repo, immutable record tying that tag to the deployed content digest.

## What it does
After a successful deploy, push an annotated git tag:
- staging: `deploy/staging/<sha_short>` → `… = <digest> (commit <sha>)`
- production: `deploy/production/<sha_short>` → `… = <digest>`

A verifier then has a public, in-git link: **attested tag → recorded digest D**, and (once reproducible builds land) **rebuild commit → D** — closing most of the registry-trust gap without needing digest-in-compose.

## Implementation notes
- Uses the build's existing `digest` output / the promote `Resolve` step — no extra registry auth needed.
- Adds `contents: write` to the deploy jobs (production keeps `packages: write` for the `:production` retag).
- Annotated tags (`-fa`); idempotent on re-runs (reproducible ⇒ same sha → same digest).

## Limits (honest)
- This is an **audit/anchor improvement, not a security guarantee** — the attestation still references a tag, so it doesn't *prove* the enclave pulled D. The trustless version is anchor (c) (digest-in-compose), blocked upstream.
- Only meaningful once **reproducible builds** (Phase A #47 + Phase B) actually converge.

## Validation ⚠️
Can't run CI here. Needs a real deploy to confirm the tag push works under branch/tag protection and `GITHUB_TOKEN` `contents: write`. If tag protection blocks `deploy/*`, we'll adjust (or switch to a committed `deployments/*.json`).

### DIFF
--- a/.github/workflows/build-scribe-api.yml
+++ b/.github/workflows/build-scribe-api.yml
@@ -85,9 +85,13 @@ jobs:
     needs: build
     runs-on: ubuntu-latest
     environment: staging
+    # contents: write lets the post-deploy step push the digest-record git tag
+    permissions:
+      contents: write
     steps:
       - uses: actions/checkout@v4
-      - uses: ./.github/actions/phala-deploy
+      - id: deploy
+        uses: ./.github/actions/phala-deploy
         with:
           cvm-name: os-scribe-api-staging
           compose-file: iac/phala/docker-compose.staging.yml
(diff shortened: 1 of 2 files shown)


jakubswierczek:
Superseded by #45, which now consolidates the plan + Phase A + anchor (b) into one PR.

PR #69, opened 2026-06-04
[codex] Fix long dual-source recording stability

andrewwashuta:
## Summary

- Throttle selected-device dictation audio-level events to ~25Hz while preserving peak levels, preventing HUD waveform event backlog on long system+mic recordings.
- Prevent system-audio readiness probes from terminating the active system recorder during an ongoing capture.
- Skip focus-triggered source readiness refresh while recording/finalizing/validating.

## Root Cause

Long selected-device dictation recordings emitted an `audio_level` event for every capture buffer, which could flood the Tauri event path and leave the HUD rendering stale waveform data.

Separately, system-audio readiness checks run a permission probe that terminates existing system-audio helpers. The app could trigger that readiness check on window focus during an active recording, killing the real system recorder while microphone capture continued.

## Validation

- `git pull --ff-only origin main`
- `swiftc -typecheck src-tauri/native/mac-dictation-helper/main.swift`
- `cargo test --manifest-path src-tauri/Cargo.toml commands::tests::`
- `pnpm test -- app-shortcut`
- `pnpm exec tsc --noEmit`

### DIFF
--- a/src-tauri/native/mac-dictation-helper/main.swift
+++ b/src-tauri/native/mac-dictation-helper/main.swift
@@ -949,6 +949,17 @@ final class SelectedDeviceRecorder: NSObject, AVCaptureAudioDataOutputSampleBuff
     private var finishHandler: ((Error?) -> Void)?
     private let failureHandler: (Error) -> Void
     private let levelHandler: (Float) -> Void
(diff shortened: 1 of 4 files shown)


andrewwashuta:
@greptile-app

PR #71, opened 2026-06-05
Dictation HUD + recorder waveform: amplitude-lens & low-latency metering

andrewwashuta:
Reworks the dictation HUD waveform (and carries it to the note recorder) for a more responsive, tactile feel — and fixes the production signal path so it actually reads that way in the app, not just in the tuning playground.

## What changed

**Waveform model (amplitude-lens).** Replaced the per-band/literal approach with a symmetric centre-peaked lens driven by one loudness envelope: flat idle row with a single travelling pulse (timing signifier), a centre that physically bounces per syllable with a quick die-down, and a very subtle speech-coupled shimmer that carries across at peak. Shared engine (`withWaveLayers`, lens weights, ballistics, `LIVE_WAVE_OPTIONS`) so the HUD and recorder move identically.

**Signal path (the responsiveness fix).** The dictation helper now meters a peak-biased `0.8·peak + 0.2·rms` blend instead of the time-smoothed `averagePower`, at 50Hz. Auto-detect records through the low-latency `AVCaptureAudioDataOutput` capture path (live sample buffers, instantaneous peak — no peak-hold lag), with **format-aware** reading (32-bit float vs Int16) and main's peak-coalescing + flush-on-stop to avoid IPC-flood freezes.

**Recorder waveform.** Polls status at ~20Hz (was 10Hz) and coalesces the peak over the poll window, so it tracks speech as snappily as the HUD. Frontend-only — **mic + system dual-source capture/transcription untouched**.

**HUD stop-braille fix.** Ignores a stray/flushed `audio_level` once past listening, so the transcribing braille starts the instant you stop instead of the pill looking stuck until the paste lands.

**Tuning playground** (`playground/waveform-playground.html`) models the exact production signal (blend source, 50Hz) so tuning transfers faithfully.

## Verification
- `tsc` clean · `swiftc -typecheck` (helper) clean · 124 vitest tests pass
- Manually verified in-app: HUD snappiness, accurate die-down, stop→braille flow, recorder responsiveness (mic + system)

## Test plan
- [ ] Dictation HUD tracks speech and dies down immediately on Auto-detect
- [ ] Stop (button + hotkey) → braille shows immediately through transcription → paste
- [ ] Recorder waveform feels snappy while recording mic + system, and still transcribes both


andrewwashuta:
@greptile-apps

### Re: `slice(-4)` "can miss the loudest transient in a 50 ms poll window" (`Waveform.tsx`)

Looked into this closely and **kept a windowed max rather than `Math.max(...recent)`** — sharing the reasoning since it's load-bearing for the feel this PR is about.

The suggestion assumes `recentPeaks` *is* the 50 ms poll window. It isn't: it's a **fixed 24-entry deque** in `capture.rs` (`recent_peaks.len() == 24`), and we don't set a cpal `buffer_size`, so callbacks run at the default (~512 frames ≈ **~11 ms**). That makes the deque span **~260 ms**, not 50 ms.

So `Math.max(...recent)` over the whole deque holds the loudest sample for ~260 ms after you stop talking → the bars sit high and ease down slowly. That's precisely the **mushy, "unsure of itself" die-down this PR set out to fix** (the original complaint was the dictation HUD's peak-*hold* meter; the recorder change mirrors the HUD's *immediate* die-down). Maxing the full deque would quietly regress it.

**What's in `b5d0d2a` instead:** the windowed max is now an explicit, documented constant rather than a magic `4`:

```ts
const POLL_WINDOW_PEAKS = 6; // ~the 50ms poll window at the default ~11ms buffer, + headroom
const raw = recent.length > 0 ? Math.max(...recent.slice(-POLL_WINDOW_PEAKS)) : level.peak;
```

`6` covers the 50 ms poll (≈4–5 callbacks at the default buffer, +1 for smaller buffers), so it catches the early-window transient the comment is worried about, **without** the long peak-hold.

**The fully-correct version** (exact poll-window coalescing regardless of buffer size) is to have Rust **drain** `recent_peaks` on each status read — return only the peaks since the last poll — then `Math.max(...recent)` is exact with zero hold. I deliberately deferred that to a follow-up rather than land it here: it also has to drain the **dual-source system-capture** deque to stay consistent, and it changes live-recorder behaviour I can't exercise from CI. Happy to open that as a separate issue/PR if we want the buffer-size-independent guarantee.

andrewwashuta:
@greptile-apps

PR #86, opened 2026-06-08
[codex] Recover dictation auto-detect recording

junhohong:
## Summary

- Restore legacy Auto-detect dictation recording behavior by keeping `AVAudioRecorder` for the default mic path while preserving the custom capture path for explicitly selected microphones.
- Surface silent dictation terminal failures as a brief `Nothing recorded` HUD state instead of disappearing immediately.
- Add a redacted local dictation event log so future dropped dictations can be diagnosed without storing transcript text.
- Trim proxied agent chat messages before sending to Scribe API while preserving leading instructions, recent context, and assistant/tool message groups.

## Root Cause

The June 5 helper change made Auto-detect/default mic resolve through `AVCaptureDevice.default(for: .audio)` and the custom `SelectedDeviceRecorder` path. For users with no pinned microphone, that changed the actual recording pipeline from the older high-level `AVAudioRecorder` path to a custom capture-session writer. Recent lost dictation symptoms lined up with low recorded levels and `no_speech` transcription responses.

The app also hid silent transcription failures immediately, which made failed recordings look like they completed and vanished.

Separately, the local agent proxy could still send too many messages to Scribe API even after server-side limits were relaxed/deployed elsewhere, so the desktop app now trims defensively before proxying.

## Validation

- `cargo test dictation::tests --manifest-path src-tauri/Cargo.toml`
- `cargo test scribe_api::tests --manifest-path src-tauri/Cargo.toml`
- `pnpm exec vitest run src/test/hud-meeting.test.ts`
- `pnpm run lint`
- `git diff --check`

### DIFF
--- a/src-tauri/native/mac-dictation-helper/main.swift
+++ b/src-tauri/native/mac-dictation-helper/main.swift
@@ -1318,15 +1318,11 @@ final class DictationController {
         resetRecordingState()
 
         let nextRecordingURL = temporaryRecordingURL()
-        // Auto-detect now resolves to the system default input and records through
-        // the same low-latency capture path as a manually-selected mic, so it's
-        // just as snappy (instantaneous peak, accurate die-down). The
-        // AVAudioRecorder path below is only a fallback for when no capture device
-        // is available at all.
-        let captureDevice =
-            microphoneDevice(for: preferredMicrophoneID) ?? AVCaptureDevice.default(for: .audio)
-        if let captureDevice {
-            startSelectedDeviceRecording(device: captureDevice, url: nextRecordingURL)
+        // Preserve the legacy Auto-detect behavior: AVAudioRecorder delegates
+        // default-input selection and audio processing to macOS. The custom
+        // capture path is still used when the user explicitly pins a microphone.
+        if let selectedDevice = microphoneDevice(for: preferredMicrophoneID) {
+            startSelectedDeviceRecording(device: selectedDevice, url: nextRecordingURL)
             return
         }
 
--- a/src-tauri/src/dictation.rs
+++ b/src-tauri/src/dictation.rs
@@ -7,6 +7,7 @@ use crate::scribe_api::{
     cleanup_text, dictate_transcribe, DictateCleanupRequestParams, DictateTranscribeRequest,
     TranscriptionProviderResult,
 };
+use chrono::Utc;
 use serde::{Deserialize, Deserializer, Serialize};
 use std::{
     fs,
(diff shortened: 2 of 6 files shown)


andrewwashuta:
@junhohong    So the one thing this is what drives me nuts about the auto detect is if you look at the waveform actually and I don't know if there's something else we should be doing, but basically what Whisper does is they have their the microphone usually is set to the built built-in mic, I think, which is recommended. However, um they also have an auto detect option, and then they have all the other options as well, right? So like it would be, you know, whether you have an input mic, whether you have like a display mic or something like that. So what's happening on ours, in my opinion, is that it's almost getting averaged out. And I yeah, so I don't know like what the situation is or how to solve that. So it takes a while for the die down of this waveform to actually like function. Um so I don't again, I don't know if auto detect is the right move or if there's like some way to have a fallback, but can we try and mirror this off of whispers? And also you have a couple of bugs to address, I believe, in here, unless you've already addressed them, but it looks like Greptile came back with some stuff. So um yeah, if we can figure out a way to fix the die down, which is what I had done the other week, but also like still retain recordings or something like that, or I I yeah, I don't know what the solution is, but like when I stop speaking right now, it's like the die down of this waveform is slow because it is recording from like I think multiple sources and averaging them out. So just want some way to solve that if possible.

andrewwashuta:
@junhohong:


### Why the waveform “averaged out” again

This is a genuine coupling, not a regression to undo:

Snappy waveform needs raw, low-latency samples — which only the AVCaptureAudioDataOutput capture path gives.
Reliable Auto-detect recording needs AVAudioRecorder — which Jun correctly restored in #86 (the capture path was recording low levels → no_speech for default-mic users).
The catch: AVAudioRecorder only exposes its built-in meters (averagePower/peakPower), and those are smoothed + lag behind the encoder. No amount of blending fixes that — there are no raw samples to read. So on Auto-detect, you get reliable recording but a mushy meter. They can’t both come from one source here.

### What you can do right now (zero risk)

Pin your mic in dictation settings instead of Auto-detect. Jun explicitly kept the capture path for pinned mics (#86), and that path still has all our snappy, format-aware, coalesced metering. So a pinned mic = snappy waveform today, on a recording path Jun already deemed reliable. If you mostly dictate on one machine, this is the immediate fix.

### The real fix for Auto-detect (needs your call)

Decouple metering from recording: keep AVAudioRecorder writing the file (Jun’s reliable path, untouched), and add a separate AVAudioEngine input tap purely for the waveform meter — raw samples → snappy, while the recording is never touched.

I’d build it safely: behind a flag, started only after recording is confirmed, and wrapped so that if the tap fails it silently falls back to today’s AVAudioRecorder meter — recording can’t break in the failure case.

The one honest caveat: it adds a second input-device client, and on rare hardware that could perturb capture — the exact thing #86 fixed. It can’t be fully proven without on-device testing, which is why I didn’t just ship it onto your branch blind. I’d want you to test recording reliability before it merges, with the flag as the kill-switch.

Want me to implement the flagged metering tap now so you can test it on-device? Or are you good with pin-your-mic for the moment? I’ll do whichever — just didn’t want to gamble your recording reliability without a heads-up.

PR #107, opened 2026-06-09
[codex] Fix auto-detect dictation metering

andrewwashuta:
## Summary

Fixes the Auto-detect dictation waveform so it behaves like the pinned microphone path while keeping the reliable Auto-detect recording path intact.

## What changed

- Kept Auto-detect recording on `AVAudioRecorder` for the reliability restored in #86.
- Added Auto-detect waveform metering through `AVCaptureDevice.default(for: .audio)` and `AVCaptureAudioDataOutput`, using the same raw sample-buffer level parser as pinned microphones.
- Extracted the Swift sample-buffer level calculation into a shared helper used by both pinned and Auto-detect capture metering.
- Preserved the existing HUD ambient gate/whisper shaping so pinned-mic HUD users do not see extra ambient activity from the recorder waveform curve.
- Left a kill switch: `OS_SCRIBE_DICTATION_RAW_METER=0` falls back to `AVAudioRecorder` metering.

## Impact

Auto-detect dictation should now have snappy waveform rise and die-down comparable to selecting the MacBook microphone directly, without changing the audio file recording path or the HUD's existing ambient-noise behavior for pinned microphones.

## Validation

- Direct `swiftc` compile of `src-tauri/native/mac-dictation-helper/main.swift`
- `pnpm exec vitest run src/test/recorder.test.tsx src/test/hud-meeting.test.ts`
- `pnpm run lint`
- `cargo test --manifest-path src-tauri/Cargo.toml dictation`

Also manually tested Auto-detect and pinned MacBook microphone behavior in the app; Auto-detect now appears basically comparable to the direct MacBook mic path.


andrewwashuta:
@greptile-apps

andrewwashuta:
@greptile-apps

PR #117, opened 2026-06-09
Defer recording consent to a non-blocking reminder

andrewwashuta:
Clicking record no longer stalls behind a blocking "Cancel / Continue" consent dialog — recording now starts immediately, and the consent message instead surfaces as a non-blocking reminder that fades in over the recorder once recording begins (auto-dismissing, with a Dismiss link). The reminder reuses the shared `InlineNotice` surface (the same component as the interrupted-recording notice), now with an optional eyebrow, and is styled to match the "Generating notes…" pill's padding, height, radius, and muted surface so the two status messages read as one family. It's centered over the recorder pill in the canvas, fades in and out via `AnimatePresence`, and uses the app's link color (`var(--primary)`) for the Dismiss action. Also fixes a recorder pill that collapsed to a thin capsule once nested in the new consent dock (a `calc(100%…)` → `calc(100vw…)` width fix). Tests in `note-editor.test.tsx` are updated to cover immediate recording start, the delayed/centered reminder, and dismissal.

### DIFF
--- a/src/components/note-editor/NoteEditor.tsx
+++ b/src/components/note-editor/NoteEditor.tsx
@@ -184,20 +184,29 @@ export function NoteEditor({
         : recordingActive && !prev.recording;
 
     consentEdgeRef.current = { noteId: note.id, recording: recordingActive };
+    // Undo the ref mutation on cleanup so StrictMode's double-invoke replays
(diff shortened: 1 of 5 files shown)


andrewwashuta:
@greptile-apps

PR #128, opened 2026-06-09
Add post-validation FLAC compression for saved recordings

### LINKED ISSUE #7: Add audio compression and storage format strategy
## Problem

The app currently stores local recording artifacts as WAV files. WAV is simple and reliable, but long meetings can consume significant disk space, especially when recording both microphone and system audio.

We need a compression strategy that reduces storage usage without compromising transcription reliability or recovery.

## Goals

- Reduce local storage usage for long recordings.
- Keep recording capture reliable.
- Keep retry/recovery possible from local audio.
- Preserve enough quality for transcription.
- Avoid adding fragile format conversion into the live capture path.

## Format comparison

| Format | Compression | Quality | Typical size vs WAV | Transcription suitability | Pros | Cons |
| --- | --- | --- | --- | --- | --- | --- |
| WAV PCM | None | Lossless | 100% | Excellent | Simple, robust, easy validation, easy turn extraction | Very large |
| FLAC | Lossless | Lossless | ~40-70% of WAV | Excellent | Smaller without quality loss, good archival option | Needs encoding/decoding support; less native than WAV in some paths |
| Opus | Lossy | Very good for speech | ~5-15% of WAV | Very good | Excellent speech compression, low bitrate, efficient for meetings | More conversion complexity; container/support decisions needed |
| AAC/M4A | Lossy | Good | ~10-25% of WAV | Good | Native macOS support, broadly compatible | Less ideal than Opus for speech efficiency; lossy artifacts possible |
| MP3 | Lossy | Good at higher bitrates | ~10-25% of WAV | Good | Universal compatibility | Older codec, less efficient, not ideal for archival |

### MVP-safe approach

Keep recording capture as WAV, then compress after validation.

1. Record microphone/system as WAV.
2. Validate the WAV artifacts.
3. Transcribe from WAV or validated turn WAVs.
4. After successful validation, create compressed archival copies.
5. Keep or delete original WAV based on a retention policy.

This avoids placing compression on the critical recording path.

### Preferred storage policy

| Stage | Artifact | Purpose | Retention |
| --- | --- | --- | --- |
| During recording | `.partial.wav` | Reliable capture/recovery | Until finalized |
| After validation | `.wav` | Source of truth for immediate processing/retry | Keep until compressed copy is verified |
| After compression | `.flac` or `.m4a` / `.opus` | Long-term local storage | Keep |
| Temporary processing | turn `.wav` files | Provider upload / extraction | Delete after processing |

### Option A: FLAC archival

Use FLAC for long-term storage after validation.

- Best when reliability and quality are top priority.
- Keeps audio lossless.
- Reduces storage meaningfully without transcription quality risk.
- Requires adding encode/decode support and validating decoded duration/signal.

### Option B: Opus speech archival

Use Opus for large space savings.

- Best when storage is the priority.
- Strong fit for speech and meetings.
- Needs careful provider compatibility and local decode path.
- Should keep WAV until Opus file is verified.

### Option C: AAC/M4A macOS-native archival

Use AAC/M4A through macOS-native tooling.

- Good compatibility on macOS.
- Smaller than WAV.
- Easier native playback story.
- Lossy, so choose conservative bitrate for transcription recovery.

## Proposed solution

Start with post-validation compression, not live compression.

1. Add an audio storage setting with default `WAV only`.
2. Add `Compressed after validation` mode behind config or preferences.
3. Implement one compression backend first, preferably FLAC for quality or AAC/M4A for native macOS support.
4. Store compressed artifact metadata in SQLite: format, original checksum, compressed checksum, duration, size, status.
5. Validate compressed artifacts:
   - file exists
   - non-zero size
   - readable audio
   - duration close to original
   - non-silent signal preserved
6. Only delete the WAV source after compressed validation passes and policy allows deletion.
7. Ensure retry can decode/use compressed audio or regenerate temporary WAV segments from it.

## Acceptance criteria

- Long recordings can be compressed after successful validation.
- Compression never runs before capture is finalized and validated.
- The app records original and compressed artifact metadata.
- The app reports original size, compressed size, and compression ratio.
- Retry still works after compression.
- The user can keep WAV originals or allow deletion after compressed validation.
- Compression failure does not mark the recording/transcription as failed.

KaoRz:
Closes #7

## What

Implements the MVP-safe storage strategy from the issue: capture keeps recording plain WAV, and compression happens **after** finalization + validation, never on the live capture path.

- **Settings (default unchanged):** new audio storage preference persisted in `audio-storage-settings.json`, default `WAV only`. Settings → Audio gains two rows: "Compress saved recordings" and, when enabled, "Keep WAV originals".
- **FLAC backend (Option A from the issue):** streaming encode via `flacenc` (no full-file memory load for long meetings), decode via `claxon`.
- **Validation before anything else:** the archive must exist, be non-empty, decode fully, and match the WAV source **sample-for-sample** (FLAC is lossless, so exact equality is required, which subsumes the duration/non-silence checks). An archive that fails validation is deleted on the spot.
- **Storage policy:** the WAV original is deleted only when the user disabled "Keep WAV originals" *and* the archive passed validation. Deletion is recorded (`original_removed_at`).
- **Metadata in SQLite:** `compressed_path/format/size/checksum`, `compression_status`, `compression_error` on `audio_artifacts`; original size, compressed size, and compression ratio are exposed on `AudioArtifactDto` and recorded in an `audio_compression` checkpoint per source.
- **Retry still works after compression:** `materialize_retry_audio` restores the WAV from the FLAC archive into its original location before reprocessing.
- **Failure isolation:** compression errors are recorded per artifact and never mark the recording or note as failed.

## How it sequences

Compression is enqueued on the existing per-note processing queue right after `finish_recording` schedules transcription, so it only runs once transcription/generation for that recording has finished with the WAV files, and it never competes for CPU during processing.

## Tests

- `src-tauri/src/audio/compression.rs`: round-trip (mono + stereo), corrupted/truncated/empty/mismatched archive rejection, non-16-bit-PCM rejection, ratio math.
- `src-tauri/tests/compression.rs`: end-to-end against a real SQLite + recordings dir — compress-and-keep policy, delete-after-validation policy, retry restore from FLAC (sample-exact), failure keeps WAV + note status untouched, no re-compression of already-archived artifacts.
- `src/test/app-settings.test.tsx`: settings rows render, toggle, and persist through the new commands.
- Full gate: `cargo test` (118), `pnpm test` (211), `pnpm lint`, `pnpm build` all green; `cargo clippy` introduces no new warnings; prettier failures on main are pre-existing and untouched.

## Notes / follow-ups

- Sizes and ratio are surfaced via DTO + checkpoint; a small UI affordance showing "12.3 MB → 4.1 MB" on the note could be layered on top later.
- A concurrent manual retry during the archive-deletion window is self-healing (the next retry restores from FLAC), mirroring the existing retry semantics.

### DIFF
--- a/src-tauri/src/audio/compression.rs
+++ b/src-tauri/src/audio/compression.rs
@@ -1,0 +1,440 @@
+//! Post-validation FLAC compression for saved recording artifacts.
+//!
+//! Compression is archival only: capture always records WAV, and a FLAC copy
+//! is produced after the WAV has been finalized and validated. FLAC is
+//! lossless, so validation can require the decoded samples to match the WAV
+//! source exactly before the original is ever eligible for deletion.
+
+use crate::domain::types::AppError;
+use flacenc::component::BitRepr;
+use flacenc::error::Verify;
+use flacenc::source::{Fill, Source};
+use hound::{SampleFormat, WavReader, WavSpec, WavWriter};
+use std::io::BufReader;
+use std::path::{Path, PathBuf};
+
+pub const FLAC_FORMAT: &str = "flac";
+
+#[derive(Debug, Clone, PartialEq)]
+pub struct CompressionOutcome {
+    pub output_path: PathBuf,
+    pub format: String,
+    pub original_size_bytes: i64,
+    pub compressed_size_bytes: i64,
+    pub compressed_checksum: String,
+    pub duration_ms: i64,
+}
+
+/// Streams 16-bit PCM WAV samples into the FLAC encoder without loading the
+/// whole recording into memory (long meetings can be hundreds of MB).
+struct WavSource {
+    reader: WavReader<BufReader<std::fs::File>>,
+    spec: WavSpec,
+    buffer: Vec<i32>,
+    failed: bool,
+}
+
+impl WavSource {
+    fn open(path: &Path) -> Result<Self, AppError> {
+        let reader = WavReader::open(path)
+            .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+        let spec = reader.spec();
+        ensure_compressible_spec(spec)?;
+        Ok(Self {
+            reader,
+            spec,
+            buffer: Vec::new(),
+            failed: false,
+        })
+    }
+}
+
+impl Source for WavSource {
+    fn channels(&self) -> usize {
+        self.spec.channels.max(1) as usize
+    }
+
+    fn bits_per_sample(&self) -> usize {
+        16
+    }
+
+    fn sample_rate(&self) -> usize {
+        self.spec.sample_rate as usize
+    }
+
+    fn read_samples<F: Fill>(
+        &mut self,
+        block_size: usize,
+        dest: &mut F,
+    ) -> Result<usize, flacenc::error::SourceError> {
+        let channels = self.channels();
+        self.buffer.clear();
+        for sample in self
+            .reader
+            .samples::<i16>()
+            .take(block_size.saturating_mul(channels))
+        {
+            match sample {
+                Ok(sample) => self.buffer.push(sample as i32),
+                Err(_) => {
+                    // A torn trailing frame in an otherwise-finalized WAV
+                    // would corrupt the archive; surface it instead.
+                    self.failed = true;
+                    return Err(flacenc::error::SourceError::from_unknown());
+                }
+            }
+        }
+        // The encoder expects whole frames; drop a torn trailing frame.
+        self.buffer
+            .truncate((self.buffer.len() / channels) * channels);
+        dest.fill_interleaved(&self.buffer)?;
+        Ok(self.buffer.len() / channels)
+    }
+}
+
+/// Encodes a finalized 16-bit PCM WAV into a FLAC archive copy.
+pub fn compress_wav_to_flac(
+    input_path: &Path,
+    output_path: &Path,
+) -> Result<CompressionOutcome, AppError> {
+    let original_size_bytes = file_size(input_path)?;
+    let mut source = WavSource::open(input_path)?;
+    let config = flacenc::config::Encoder::default()
+        .into_verified()
+        .map_err(|(_, error)| AppError::new("audio_compression_failed", error.to_string()))?;
+    let stream = flacenc::encode_with_fixed_block_size(&config, &mut source, config.block_size)
+        .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    if source.failed {
+        return Err(AppError::new(
+            "audio_compression_failed",
+            "Source WAV could not be fully read while encoding.",
+        ));
+    }
+    let mut sink = flacenc::bitsink::ByteSink::new();
+    stream
+        .write(&mut sink)
+        .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    if let Some(parent) = output_path.parent() {
+        std::fs::create_dir_all(parent)
+            .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    }
+    std::fs::write(output_path, sink.as_slice())
+        .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    let compressed_size_bytes = file_size(output_path)?;
+    let compressed_checksum = crate::audio::validation::checksum_file(output_path)
+        .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    Ok(CompressionOutcome {
+        output_path: output_path.to_path_buf(),
+        format: FLAC_FORMAT.to_string(),
+        original_size_bytes,
+        compressed_size_bytes,
+        compressed_checksum,
+        duration_ms: flac_duration_ms(output_path)?,
+    })
+}
+
+/// Verifies a FLAC archive decodes to exactly the samples of its WAV source.
+/// FLAC is lossless, so anything short of an exact match (spec, length, every
+/// sample) means the archive cannot replace the original.
+pub fn validate_flac_matches_wav(flac_path: &Path, wav_path: &Path) -> Result<(), AppError> {
+    let size = file_size(flac_path)?;
+    if size == 0 {
+        return Err(AppError::new(
+            "audio_compression_invalid",
+            "Compressed audio file is empty.",
+        ));
+    }
+    let mut flac = claxon::FlacReader::open(flac_path)
+        .map_err(|error| AppError::new("audio_compression_invalid", error.to_string()))?;
+    let info = flac.streaminfo();
+    let mut wav = WavReader::open(wav_path)
+        .map_err(|error| AppError::new("audio_compression_invalid", error.to_string()))?;
+    let spec = wav.spec();
+    ensure_compressible_spec(spec)?;
+    if info.sample_rate != spec.sample_rate
+        || info.channels != spec.channels as u32
+        || info.bits_per_sample != 16
+    {
+        return Err(AppError::new(
+            "audio_compression_invalid",
+            "Compressed audio format does not match the WAV source.",
+        ));
+    }
+    let mut wav_samples = wav.samples::<i16>();
+    let mut compared = 0_u64;
+    let mut non_silent_preserved = false;
+    for flac_sample in flac.samples() {
+        let flac_sample = flac_sample
+            .map_err(|error| AppError::new("audio_compression_invalid", error.to_string()))?;
+        let Some(wav_sample) = wav_samples.next() else {
+            return Err(AppError::new(
+                "audio_compression_invalid",
+                "Compressed audio is longer than the WAV source.",
+            ));
+        };
+        let wav_sample = wav_sample
+            .map_err(|error| AppError::new("audio_compression_invalid", error.to_string()))?;
+        if flac_sample != wav_sample as i32 {
+            return Err(AppError::new(
+                "audio_compression_invalid",
+                "Compressed audio does not match the WAV source signal.",
+            ));
+        }
+        if wav_sample != 0 {
+            non_silent_preserved = true;
+        }
+        compared += 1;
+    }
+    if wav_samples.next().is_some() {
+        return Err(AppError::new(
+            "audio_compression_invalid",
+            "Compressed audio is shorter than the WAV source.",
+        ));
+    }
+    if compared == 0 {
+        return Err(AppError::new(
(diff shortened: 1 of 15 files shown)


junhohong:
@greptile review

src-tauri/src/audio/compression.rs
```diff
@@ -0,0 +1,440 @@
...
+    }
+    let mut sink = flacenc::bitsink::ByteSink::new();
+    stream
+        .write(&mut sink)
+        .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    if let Some(parent) = output_path.parent() {
+        std::fs::create_dir_all(parent)
+            .map_err(|error| AppError::new("audio_compression_failed", error.to_string()))?;
+    }
+    std::fs::write(output_path, sink.as_slice())
```
junhohong, line 121: Confirming the memory concern from the bot review, with corrected math — it's worse than stated. flacenc 0.5.1's `encode_with_fixed_block_size` does read the source incrementally, but it accumulates **every encoded frame** in the returned `Stream { frames: Vec<Frame> }`, and each LPC subframe holds its residuals as `Vec<u32>` quotients + remainders — ~8 bytes per sample per channel, i.e. ~4× the 16-bit PCM data. Then `ByteSink` adds the full serialized FLAC on top before `fs::write`.

For a 2-hour 48 kHz stereo recording (WAV ≈ 1.4 GB): `Stream` ≈ 5.5 GB + `ByteSink` ≈ 0.7 GB — enough to OOM an 8 GB machine. The doc comment on `WavSource` ("without loading the whole recording into memory") is true of the *input* but contradicted by the output path.

Two fixes needed here:
1. **Constant-memory encode:** flacenc exposes `encode_fixed_size_frame` — encode frame-by-frame, write each frame through a `BufWriter<File>`, then seek back and patch STREAMINFO (total samples + MD5 are at fixed offsets).
2. **Atomic output:** write to `<name>.flac.tmp` and rename after success. Right now a crash between `fs::write` and `record_artifact_compression` leaves an orphan `.flac` with no DB row — it's never re-validated, never re-attempted, and `delete_note` won't remove it (it only deletes DB-recorded paths).

src-tauri/src/commands.rs
```diff
@@ -875,6 +880,15 @@ async fn finish_recording_session(
         }
         ticket.finish();
     });
+    // Archival compression (when enabled) queues behind the processing job we
+    // just registered, so it runs only after transcription has finished with
+    // the WAV originals.
+    crate::audio_storage::schedule_session_compression(
```
junhohong, line 886: The comment above says compression "queues behind the processing job we just registered", but I don't think the queue guarantees that. `processing_queue::enqueue` hands every job a clone of the *same* per-note `AsyncMutex` — there's no FIFO chain by enqueue order. Ordering is decided by which spawned task reaches `.lock().await` first, and both the transcription task (spawned above) and the compression task (spawned inside `schedule_session_compression`, which also calls `enqueue` *inside* the spawn) are freshly spawned tasks racing on a multi-threaded runtime.

If compression wins the race with `keep_wav_after_compression: false`, it deletes the WAV before `process_saved_source_audio` reads it → the fresh recording fails to transcribe (recoverable via retry + FLAC restore, but user-visible).

Deterministic fix: enqueue the compression ticket synchronously here (right after the transcription `enqueue` at the top of this function) so the mutex is acquired in a known order — or simpler, pass both tickets to a single spawned task that runs transcription then compression sequentially.

src-tauri/src/audio_storage.rs
```diff
@@ -0,0 +1,374 @@
...
+    paths: AppPaths,
+    note_id: String,
+    session_id: String,
+) {
+    let settings = configured_audio_storage_settings();
+    if settings.mode != AudioStorageMode::CompressedAfterValidation {
+        return;
+    }
+    tokio::spawn(async move {
+        let (ticket, _depth) = processing_queue::enqueue(&note_id);
```
junhohong, line 170: Scheduling compression only at `finish_recording` time leaves several holes:

- App quits/crashes before this queued task runs → `compression_status` stays NULL forever and the session is never compressed. There's no startup sweep, and `recover_recording` doesn't schedule compression either — so users who enable the setting only get it for recordings finished while the app stays alive.
- Crash after FLAC+DB record but before WAV removal (`keep_wav=false`) → the WAV is kept forever; `compressible_artifacts_for_session` filters `compression_status IS NULL` so removal is never re-attempted.

A startup/idle sweep over "validated, processing-complete, `compression_status IS NULL`" artifacts would close all of these at once, and would also remove the need for the fragile ordering dance with the transcription task (see my comment in `commands.rs`).

Minor, same function: this ticket shares the `pending` counter that drives the UI "queued recordings" chip via `queued_behind()`, so a note with compression waiting shows a phantom queued recording.

src-tauri/src/db/repositories.rs
```diff
@@ -1537,44 +1538,102 @@ impl Repositories {
...
+    ) -> Result<Vec<CompressibleArtifact>, sqlx::Error> {
         let rows = sqlx::query(
-            "SELECT id, source, format, duration_ms, size_bytes, checksum, created_at
+            "SELECT id, source, path
              FROM audio_artifacts
-             WHERE note_id = ? AND status = 'valid'
-             ORDER BY created_at DESC",
+             WHERE recording_session_id = ?
+               AND status = 'valid'
+               AND compression_status IS NULL
```
junhohong, line 1562: `compression_status IS NULL` means a *failed* compression is permanent: a transient failure (disk full, file briefly locked) writes `status='failed'` and the artifact is excluded from this query forever, with no retry path and no UI surface. Consider including `'failed'` rows with a bounded attempt count, or at least making this visible somewhere.

src-tauri/src/domain/processing.rs
```diff
@@ -782,6 +783,69 @@ pub async fn retry_from_saved_audio(
...
+    if !flac_path.exists() {
+        return None;
+    }
+    let restored = wav_path.clone();
+    match tokio::task::spawn_blocking(move || {
+        crate::audio::compression::decode_flac_to_wav(&flac_path, &restored)
+    })
+    .await
+    {
+        Ok(Ok(path)) => {
```
junhohong, line 825: Two follow-ups on the restore path:

1. After a retry restores the WAV, nothing ever re-archives or re-removes it — `compression_status` is already `succeeded` so the artifact never re-enters the compressible query, and `original_removed_at` now points at a file that exists again. One retry permanently leaves the user storing WAV + FLAC ≈ 155% of the original — the opposite of the feature's goal. Re-scheduling removal (or clearing the flag) after a successful retry would fix it.
2. `compressed_checksum` is stored but never read — this restore is the one moment where verifying it matters (bit-rot detection currently rests only on claxon's frame CRCs). Either check it here before decoding, or drop the column.

junhohong (commented):
I went through this with fresh eyes (independent of the Greptile pass), including the flacenc 0.5.1 internals. Overall the validation/restore logic is carefully done and the constraints from #7 are honored — capture path untouched, default unchanged, sample-exact validation, delete-only-after-validation. But there are a few structural problems worth fixing before merge, and one bigger question about whether this is the right point in the design space.

## On the existing Greptile comments

1. **Memory buffering (P1)** — confirmed, and it's actually *worse* than stated. See my inline comment: peak RAM is ~4× the PCM data (not ~2× the FLAC size), which is multi-GB for a long meeting. The input side is fine (the WAV source and validation genuinely stream); it's the encoder output that's fully materialized.
2. **`non_silent_preserved` dead code (P2)** — confirmed. It's also logically redundant (the sample-exact loop subsumes it, and fully-silent sources must stay valid), so the right fix is to delete the accumulation rather than enforce it.
3. **Panic skips `ticket.finish()` (P2)** — real but overstated. It's byte-for-byte the same pattern as the existing transcription path (`commands.rs:826-882`), tokio's mutex doesn't poison so later jobs still run, and the CPU-heavy work is in `spawn_blocking` whose panics surface as errors. What actually leaks is the `pending` counter → a permanently inflated "queued recordings" chip. A `Drop`-based ticket guard would fix both call sites; fine as a follow-up.

## Additional findings not covered inline

- **`retry_processing` bypasses the per-note queue** (`commands.rs:944-952`, pre-existing) — but this PR makes the bypass dangerous: a retry can resolve the WAV path while the queued compression job concurrently deletes that WAV (`keep_wav=false`). Routing retry through `processing_queue::enqueue` like `finish_recording` does would close it.
- **Compression jobs inflate the user-visible queue count.** The archival ticket shares the `pending` counter that drives `queued_behind()` (`commands.rs:106`), so while compression waits behind transcription the note shows a phantom queued recording.
- Nits: `CompressionOutcome.duration_ms` is only read by tests and re-opens the file to compute; settings are written with a bare `fs::write` (no temp+rename), and a corrupt file silently falls back to WavOnly.

Verified-good, for the record: encode/validate/decode all run under `spawn_blocking` (no runtime blocking); validation is genuinely streaming and sample-exact; migrations follow the additive `ensure_column` pattern; `delete_note` now also cleans `partial_path`/`compressed_path` (fixes a pre-existing leak); UI rows are sentence-case, disabled until settings load, optimistic with rollback.

## Is this the right approach?

The *shape* (post-validation, off the capture path) is right per #7. Two things I'd push on:

1. **FLAC saves ~45%, and the defaults can make things worse.** With `keep_wav_after_compression: true` (the default once compression is enabled), disk usage *grows* ~55% until the user finds the second toggle. The lossless argument is weak when the only consumer of a retry is a lossy transcription model — Opus at 32-48 kbps is 90-95% savings and would eliminate the restore machinery entirely if the pipeline can ingest it. If lossless stays, FLAC's built-in STREAMINFO MD5 already covers decode fidelity — comparing that against an MD5 computed during the encode pass gives the same guarantee as the second decode-and-compare pass at half the I/O.
2. **Schedule-at-finish is the root cause of most issues here** (ordering race, retry race, never-compressed sessions, orphan files). A startup/idle *sweep* over "validated + processing-complete + `compression_status IS NULL`" artifacts would be simpler and strictly safer — it has no ordering to get wrong, retries transient failures for free, and picks up sessions from before the feature existed. It also pairs naturally with the simplest policy of all, worth at least considering: "compress (or delete) WAVs older than N days."

None of this blocks the feature conceptually — but I'd want the memory fix, the ordering race, and atomic output writes (tmp + rename) addressed before this ships.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

junhohong:
jun (actual human review): i dont think we should even be saving recordings

PR #129, opened 2026-06-09
Transcribe dual-source recordings incrementally while still recording

### LINKED ISSUE #6: Optimize transcription latency for long recordings
## Problem

Long recordings still take too long to finish processing after the user stops recording. The current implementation validates and finalizes audio first, then detects turns, extracts segment WAVs, transcribes source lanes, and generates the final note. This is reliable, but most transcription work still starts after `Done`.

For 30-minute meetings, the user can be left waiting while the note stays in a processing state.

## Current behavior

- Audio is saved locally and validated before processing.
- Dual-source recordings are split into ordered turns after recording finalization.
- Microphone and system source lanes can run concurrently.
- Turns within each source lane are transcribed sequentially to preserve context.
- Transcripts are persisted only after the lane work completes.
- The final note is generated after all valid transcripts are assembled.

## Proposed solution

Implement an incremental transcription pipeline that starts before recording ends while preserving local audio reliability.

### Phase 1: Instrument latency

Add timing logs/checkpoints for:

- recording finalization
- audio validation
- turn detection
- turn WAV extraction
- each transcription request
- transcript persistence
- note generation
- UI polling completion

This should make slow stages visible in development logs and persisted recording checkpoints.

### Phase 2: Persist partial turn results

Persist each successful turn transcript as soon as it completes instead of waiting for every source lane to finish. The Transcription tab should be able to show partial results ordered by timing metadata while later turns are still running.

### Phase 3: Add bounded turn concurrency

Move from source-lane-only concurrency to bounded turn-level concurrency:

- default max concurrent transcription requests: 3 or 4
- preserve chronological ordering when storing/displaying results
- keep recent transcript context where available, but do not block all later turns on one slow turn
- use context from already-completed nearby turns only
- keep failed turns retryable without discarding successful turns

### Phase 4: Optional rolling chunks during recording

If post-finalization latency remains too high, change capture to emit finalized rolling chunks during recording:

- full source WAV remains the source of truth
- rolling chunks are secondary artifacts for early transcription
- chunks must have valid WAV headers before provider upload
- chunk boundaries should include padding to avoid clipped words
- final processing reconciles rolling chunk transcripts with saved full audio

## Reliability constraints

- Do not weaken audio validation.
- Do not make provider availability part of the recording capture path.
- Keep saved full audio available for retry and recovery.
- Never discard already successful turn transcripts because a later turn fails.
- Preserve `Microphone` / `System` source labels and chronological ordering.

## Acceptance criteria

- A 30-minute dual-source recording begins transcription work before final note generation starts.
- Partial transcript rows appear before every turn has completed.
- The UI remains in a clear processing state until generation is complete or failed.
- Retry can reuse saved audio and any successful persisted turn transcripts where safe.
- Development logs/checkpoints show where processing time is spent.
- Bounded concurrency does not produce duplicate, missing, or out-of-order transcript rows.

KaoRz:
Closes #6

## Context

Phases 1–3 of the issue (latency checkpoints, partial turn persistence shown live in the UI, bounded turn concurrency of 4) already landed in #60. What remained was the substantive Phase 4 — starting transcription **before the recording ends** — plus two small Phase 1/2 gaps (a UI polling completion checkpoint and test coverage for partial rows during processing). This PR implements those.

## How it works

**Live loop (dual-source recordings only).** When a microphone+system recording starts, a per-session background loop ticks every 15s:

1. `audio/live.rs` reads the in-progress `.partial.wav` files — their RIFF headers still hold placeholder sizes, so it walks the chunks and sizes the data section from the bytes actually on disk. RMS windows and interval detection are the *same code paths* as finalized-audio detection (`turns.rs` was refactored to share them), so live and final boundaries agree.
2. Turns from both sources are sorted and coalesced exactly like the final pipeline, then only turns that ended **at least 4s before the live edge** are transcribed. 4s deliberately exceeds the 2.5s transcription coalesce gap: a turn that ended that long ago can never later merge with speech arriving after the edge, which is what guarantees no clipped words at chunk boundaries.
3. Successful transcripts are persisted as provisional rows keyed by start time (chronological, collision-free per source), so the Transcription tab shows them in order while later turns are still running.

**Reconciliation at Done.** `finish_recording` signals the loop synchronously; the background processing task drains it, then final detection runs on the saved full WAVs (still the source of truth). The old exact `(source, turn_index)` transcript cache is replaced by time-range matching: persisted rows contained in a final turn (±1.2s/±1.5s boundary tolerance) are joined in order and reused when they cover ≥80% of it — anything less is re-transcribed from the saved audio. Reused text is re-persisted under final boundaries/indices and unmatched provisional rows are pruned, so the session's stored rows converge on final detection. Retry reuse (identical boundaries) matches exactly, as before.

## Reliability constraints from the issue

- Audio validation is untouched; capture never depends on the provider. Live failures are logged, counted in a `live_transcription` checkpoint, and left for the final pass (no hot-loop retries).
- Saved full audio remains available for retry; successful turn transcripts are never discarded because a later turn fails (failed live turns simply aren't persisted; reconciliation only *replaces* provisional rows with their final-keyed equivalents).
- `Microphone`/`System` labels and chronological ordering preserved; the upsert key + prune prevent duplicate or stale rows.
- Kill switch: `OS_SCRIBE_LIVE_TRANSCRIPTION=0` disables the loop entirely.

## Instrumentation

- Per-turn `live_transcription_request` source checkpoints and a session-level `live_transcription` summary.
- New `ui_polling_complete` checkpoint recorded by the frontend when polling first observes a terminal status (kind whitelist on the command).

## Tests

- `audio/live.rs`: placeholder-header parsing, partial-vs-finalized RMS window equality, segment extraction with valid headers, clamping, detection on partial audio, non-WAV rejection.
- `domain/live_transcription.rs` (fake transcriber + real SQLite): eligible turn transcribed once and persisted provisionally, near-edge turns skipped, failed turns not persisted and not re-hammered, stop flag honored.
- `domain/processing.rs`: reuse matcher — exact, drifted, coalesced-join, insufficient coverage, wrong source, out-of-range, degenerate full-source turns.
- `tests/live_reconciliation.rs`: prune removes only unmatched provisional rows; reconciliation upsert replaces same-key rows.
- Frontend: partial turn rows render while the note is `transcribing`.
- Full gate: `cargo test` (195), `pnpm test` (211), `pnpm lint`, `pnpm build` green; no new clippy warnings; prettier failures on main are pre-existing.

## Notes

- Mic-only recordings keep the existing whole-file pipeline (no turn transcripts to reuse there); scoping live work to dual-source matches the issue's 30-minute dual-source acceptance scenario.
- During the few seconds of reconciliation, the UI can briefly show final-keyed rows alongside not-yet-pruned provisional rows; prune at the end of the pass settles it.

### DIFF
--- a/src-tauri/src/audio/live.rs
+++ b/src-tauri/src/audio/live.rs
@@ -163,17 +163,65 @@ pub fn rms_windows_from_partial(path: &Path) -> Result<(PartialWavInfo, Vec<f32>
     Ok((info, windows))
 }
 
+/// RMS windows accumulated across live ticks so each tick only reads the
+/// audio appended since the previous one, instead of re-reading the whole
+/// in-progress file (which grows without bound during long meetings).
+#[derive(Debug, Default)]
+pub struct RmsWindowCache {
+    windows: Vec<f32>,
+    frames_consumed: u64,
+}
+
+/// Extends the cache with complete RMS windows from frames appended since the
+/// last tick. Only whole windows are cached: the trailing partial window
+/// changes as the file grows, and the live-edge padding keeps anything that
+/// close to the edge out of transcription anyway.
+fn extend_window_cache(
+    path: &Path,
+    info: &PartialWavInfo,
+    cache: &mut RmsWindowCache,
+) -> Result<(), AppError> {
+    let channels = info.channels.max(1) as u64;
+    let frames_per_window =
+        ((info.sample_rate.max(1) as i64 * turns::WINDOW_MS) / 1000).max(1) as u64;
+    if info.frame_count < cache.frames_consumed {
+        // The file shrank (recreated source); start over from scratch.
+        cache.windows.clear();
+        cache.frames_consumed = 0;
+    }
+    let new_windows = (info.frame_count - cache.frames_consumed) / frames_per_window;
+    if new_windows == 0 {
+        return Ok(());
+    }
+    let new_frames = new_windows * frames_per_window;
+    let samples = open_samples(path, info, cache.frames_consumed, new_frames)?;
+    let mut windows = turns::rms_windows_from_samples(
+        samples,
+        channels as usize,
+        info.sample_rate.max(1) as usize,
+    );
+    // A short read (torn tail) would yield a partial final window that the
+    // next tick must recompute; keep only the complete ones.
+    windows.truncate(new_windows as usize);
+    cache.frames_consumed += windows.len() as u64 * frames_per_window;
+    cache.windows.append(&mut windows);
+    Ok(())
+}
+
 /// Detects speech turns in an in-progress source recording using the same
-/// per-source thresholds as finalized detection. Returned turns carry the
-/// partial file as their source path; `turn_index` is left at 0.
+/// per-source thresholds as finalized detection, reading only the audio
+/// appended since the previous call with the same cache. Returned turns carry
+/// the partial file as their source path; `turn_index` is left at 0.
 pub fn detect_partial_turns(
     artifact_id: &str,
     source: &str,
     path: &Path,
+    cache: &mut RmsWindowCache,
 ) -> Result<(PartialWavInfo, Vec<AudioTurn>), AppError> {
-    let (info, windows) = rms_windows_from_partial(path)?;
+    let info = read_partial_wav_info(path)?;
+    extend_window_cache(path, &info, cache)?;
     let intervals =
-        turns::active_intervals_from_windows(&windows, turns::config_for_source(source));
+        turns::active_intervals_from_windows(&cache.windows, turns::config_for_source(source));
     let turns = intervals
         .into_iter()
         .map(|(start_ms, end_ms)| AudioTurn {
@@ -366,13 +414,64 @@ mod tests {
         write_wav(&partial, 1, 16_000, &samples);
         zero_header_sizes(&partial);
 
-        let (info, turns) =
-            detect_partial_turns("artifact", "microphone", &partial).expect("detection");
+        let (info, turns) = detect_partial_turns(
+            "artifact",
+            "microphone",
+            &partial,
+            &mut RmsWindowCache::default(),
+        )
+        .expect("detection");
         assert_eq!(info.duration_ms(), 6_000);
         assert_eq!(turns.len(), 2, "expected two distinct turns: {turns:?}");
         assert!(turns[0].start_ms < 500);
         assert!((1_000..=2_500).contains(&turns[0].end_ms));
         assert!((4_000..=5_100).contains(&turns[1].start_ms));
+
+        let _ = std::fs::remove_dir_all(dir);
+    }
+
+    #[test]
+    fn cached_detection_across_ticks_matches_detection_from_scratch() {
+        let dir = temp_dir("live-cache");
+        let partial = dir.join("partial.wav");
+        // First tick sees speech then silence; the file then grows by another
+        // silence + speech stretch before the second tick.
+        let first = tone(24_000, 12_000.0)
+            .into_iter()
+            .chain(std::iter::repeat(0).take(48_000))
+            .collect::<Vec<_>>();
+        let mut full = first.clone();
+        full.extend(std::iter::repeat(0).take(8_000));
+        full.extend(tone(24_000, 12_000.0));
+
+        write_wav(&partial, 1, 16_000, &first);
+        zero_header_sizes(&partial);
+        let mut cache = RmsWindowCache::default();
+        let (_, first_turns) = detect_partial_turns("artifact", "microphone", &partial, &mut cache)
+            .expect("first tick");
+        assert_eq!(first_turns.len(), 1);
+
+        write_wav(&partial, 1, 16_000, &full);
+        zero_header_sizes(&partial);
+        let (_, cached_turns) =
+            detect_partial_turns("artifact", "microphone", &partial, &mut cache)
+                .expect("second tick");
+        let (_, fresh_turns) = detect_partial_turns(
+            "artifact",
+            "microphone",
+            &partial,
+            &mut RmsWindowCache::default(),
+        )
+        .expect("from scratch");
+
+        let ranges = |turns: &[AudioTurn]| {
+            turns
+                .iter()
+                .map(|turn| (turn.start_ms, turn.end_ms))
+                .collect::<Vec<_>>()
+        };
+        assert_eq!(ranges(&cached_turns), ranges(&fresh_turns));
+        assert_eq!(cached_turns.len(), 2);
 
         let _ = std::fs::remove_dir_all(dir);
     }
--- a/src-tauri/src/audio/live.rs
+++ b/src-tauri/src/audio/live.rs
@@ -1,0 +1,390 @@
+//! Reading in-progress (`.partial.wav`) recordings while capture is running.
+//!
+//! A live WAV's header still carries placeholder sizes (writers only patch
+//! them on finalize), so these readers walk the RIFF chunks and size the data
+//! section from the bytes actually on disk. Live turn detection and extraction
+//! share the same window math as finalized-audio detection in
+//! [`crate::audio::turns`], so both see the same boundaries.
+
+use crate::audio::turns::{self, AudioTurn};
+use crate::domain::types::AppError;
+use hound::{SampleFormat, WavSpec, WavWriter};
+use std::{
+    fs::File,
+    io::{BufReader, Read, Seek, SeekFrom},
+    path::{Path, PathBuf},
+};
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct PartialWavInfo {
+    pub channels: u16,
+    pub sample_rate: u32,
+    pub data_offset: u64,
+    /// Whole frames currently available on disk.
+    pub frame_count: u64,
+}
+
+impl PartialWavInfo {
+    pub fn duration_ms(&self) -> i64 {
+        ((self.frame_count as i128 * 1000) / self.sample_rate.max(1) as i128) as i64
+    }
+
+    fn bytes_per_frame(&self) -> u64 {
+        self.channels.max(1) as u64 * 2
+    }
+}
+
+fn read_error(error: impl ToString) -> AppError {
+    AppError::new("audio_live_read_failed", error.to_string())
+}
+
+/// Parses the RIFF layout of a possibly still-growing 16-bit PCM WAV. The
+/// data section length is derived from the file size on disk, not from the
+/// (placeholder) chunk size in the header.
+pub fn read_partial_wav_info(path: &Path) -> Result<PartialWavInfo, AppError> {
+    let file = File::open(path).map_err(read_error)?;
+    let file_len = file.metadata().map_err(read_error)?.len();
+    let mut reader = BufReader::new(file);
+    let mut riff = [0_u8; 12];
+    reader.read_exact(&mut riff).map_err(read_error)?;
+    if &riff[0..4] != b"RIFF" || &riff[8..12] != b"WAVE" {
+        return Err(read_error("Recording is not a RIFF WAVE file."));
+    }
+
+    let mut position = 12_u64;
+    let mut format: Option<(u16, u32, u16)> = None;
+    loop {
+        let mut header = [0_u8; 8];
+        if reader.read_exact(&mut header).is_err() {
+            return Err(read_error("Recording has no audio data yet."));
+        }
+        position += 8;
+        let chunk_id = &header[0..4];
+        let stated_size = u32::from_le_bytes([header[4], header[5], header[6], header[7]]) as u64;
+        if chunk_id == b"fmt " {
+            let mut body = vec![0_u8; stated_size.min(64) as usize];
+            reader.read_exact(&mut body).map_err(read_error)?;
+            if body.len() < 16 {
+                return Err(read_error("Recording fmt chunk is too short."));
+            }
+            let audio_format = u16::from_le_bytes([body[0], body[1]]);
+            let channels = u16::from_le_bytes([body[2], body[3]]);
+            let sample_rate = u32::from_le_bytes([body[4], body[5], body[6], body[7]]);
+            let bits_per_sample = u16::from_le_bytes([body[14], body[15]]);
+            if audio_format != 1 || bits_per_sample != 16 {
+                return Err(read_error(
+                    "Only 16-bit PCM recordings support live reading.",
+                ));
+            }
+            format = Some((channels, sample_rate, bits_per_sample));
+            let skip = stated_size.saturating_sub(body.len() as u64) + (stated_size & 1);
+            if skip > 0 {
+                reader
+                    .seek(SeekFrom::Current(skip as i64))
+                    .map_err(read_error)?;
+            }
+            position += stated_size + (stated_size & 1);
+        } else if chunk_id == b"data" {
+            let Some((channels, sample_rate, _bits)) = format else {
+                return Err(read_error("Recording data chunk precedes fmt chunk."));
+            };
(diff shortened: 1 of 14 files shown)


junhohong:
@greptile review

src-tauri/src/commands.rs
```diff
@@ -819,6 +843,11 @@ async fn finish_recording_session(
     let task_session_id = finished.session_id.clone();
     let task_source_mode = finished.source_mode;
     tokio::spawn(async move {
+        // Wait for any in-flight live turn transcription to land before the
+        // final pass reads the session's persisted turn transcripts.
+        if let Some(handle) = live_handle {
```
junhohong, line 857: `drain` is a bare `handle.task.await` with no timeout, and the live loop's stop flag is only checked *between turns* and between ticks — not inside `transcribe_prepared_audio`'s sequential chunk loop. A live turn longer than the chunk size is split into N chunks transcribed sequentially, each bounded only by the 60 s reqwest timeout (plus a cleanup call per turn on non-OpenAI providers).

Worst case: the user hits Done just as a long monologue's live transcription starts → the final pass waits multiple minutes before it even begins — strictly *slower* than main for the exact long-meeting case this PR targets. (The UI isn't blocked — `finish_recording` returns immediately — but the note sits in processing.)

Suggest: check `stop` between chunks in the live path, and wrap drain in `tokio::time::timeout` (~30 s) followed by `handle.abort()`. Abort rather than detach matters: a live task that outlives the prune could upsert a provisional row *after* reconciliation, leaving a stale row in the session.

src-tauri/src/domain/processing.rs
```diff
@@ -66,15 +66,101 @@ fn source_transcript_input_from_row(row: &TranscriptDto) -> SourceTranscriptInpu
...
+/// Tolerances for reusing a transcript persisted during the live recording
+/// (or a previous retry) for a final-detection turn. Live boundaries can
+/// drift slightly from final ones because the dynamic noise floor is computed
+/// over a shorter prefix of the audio.
+const REUSE_START_TOLERANCE_MS: i64 = 1_200;
+const REUSE_END_TOLERANCE_MS: i64 = 1_500;
+/// Persisted rows must cover at least this share of the final turn's range;
+/// anything less means part of the turn was never transcribed, so it is
+/// transcribed fresh from the saved audio.
+const REUSE_MIN_COVERAGE: f64 = 0.8;
```
junhohong, line 81: The 80% threshold is *relative only* — there's no absolute cap on the uncovered span, so reuse can silently drop up to 20% of a turn's speech that was never transcribed by anyone. Concrete: final turn of 60 s; the live row (with drifted boundary — the caveat in the comment above is real, the noise floor is computed over a shorter prefix) covered only the first 50 s → 83% ≥ 80% → reused → the last 10 s of detected speech vanishes from the note even though the saved WAV contains it.

The 4 s edge guard prevents this under *identical* thresholds, but threshold drift between ticks and the final pass defeats that guarantee — and this undermines the PR's own "saved audio is source of truth" principle.

Suggest additionally requiring `turn_len − covered ≤` an absolute bound (~2 s, in line with the 2.5 s coalesce gap), falling back to fresh transcription otherwise.

src-tauri/src/domain/processing.rs
```diff
@@ -66,15 +66,101 @@ fn source_transcript_input_from_row(row: &TranscriptDto) -> SourceTranscriptInpu
...
+            covered += overlap_end - overlap_start;
+            cursor = cursor.max(overlap_end);
+        }
+    }
+    if (covered as f64) < (turn_len as f64) * REUSE_MIN_COVERAGE {
+        return None;
+    }
+    let text = candidates
+        .iter()
+        .map(|(_, _, row)| row.text.trim())
```
junhohong: The coverage cursor correctly avoids double-*counting*, but the text join concatenates **every** candidate's full text — including rows that didn't advance the cursor. `ATTEMPTED_OVERLAP = 0.6` allows two ticks to persist rows overlapping by up to 60% of the shorter one (boundary drift across ticks): e.g. rows `[0, 60000]` and `[58000, 63000]` both match a final `[0, 63000]` turn, and seconds 58–60 appear twice in the reused text.

Suggest skipping rows whose range is already covered when joining (e.g. only include rows with `row_end > cursor`, or trim rows whose `row_start < cursor` by some small tolerance).

src-tauri/src/audio/live.rs
```diff
@@ -0,0 +1,390 @@
...
+        reader: BufReader::new(file),
+        remaining_bytes: frame_count * info.bytes_per_frame(),
+    })
+}
+
+/// RMS windows over the audio currently on disk, identical to the windows
+/// finalized-audio detection computes for the same samples.
+pub fn rms_windows_from_partial(path: &Path) -> Result<(PartialWavInfo, Vec<f32>), AppError> {
+    let info = read_partial_wav_info(path)?;
+    let samples = open_samples(path, &info, 0, info.frame_count)?;
```
junhohong, line 157: Every 15 s tick re-reads and re-RMS's the **entire** partial file per source (`open_samples(path, &info, 0, info.frame_count)`), and `extract_partial_turn_wav` re-parses the header again. Over a 2-hour dual-source recording that's ~1.4 GB × 2 sources read and windowed per tick near the end, ~480 ticks total — O(duration²) I/O. The page cache softens the disk side, but CPU/battery cost on a laptop mid-meeting grows steadily as the meeting gets longer.

Since RMS windows are append-only (one f32 per 30 ms ≈ 0.5 MB/hour), caching them in the session state and reading only frames appended since the previous tick would make each tick O(new audio). The percentile threshold can still be recomputed over the cached windows each tick.

src-tauri/src/commands.rs
```diff
@@ -619,6 +634,10 @@ async fn finish_recording_session(
     finished: crate::audio::capture::FinishedRecording,
     finalization_started: Instant,
 ) -> Result<FinishRecordingResponse, AppError> {
+    // Stop the live transcription loop now that the partial files are being
+    // finalized; its in-flight work is drained before final processing reads
+    // the session's persisted turn transcripts.
+    let live_handle = crate::domain::live_transcription::signal_stop(&finished.session_id);
```
junhohong, line 649: This `signal_stop` sits *after* `finish_capture(&request.session_id)?` — if `finish_capture` errors (session mismatch, poisoned lock), the live loop is never stopped and keeps ticking every 15 s (re-reading the whole partial file each time) until app exit. The `recover_recording` discard path also never stops a live session, and it deletes files out from under a running loop.

Suggest a fire-and-forget `signal_stop` at the top of `finish_recording` before `finish_capture`, and the same in the discard path.

junhohong (commented):
Independent deep pass over this one (separately from the Greptile review), including the flow from `start_recording` → live loop → drain → reconcile → prune, and a line-by-line check of the `turns.rs` refactor against main.

First, the headline: this is the right architecture for the latency goal. Poll-detect-reconcile keeps capture provider-independent, keeps the saved WAVs as the source of truth, and degrades to exactly today's behavior when anything fails. Trusting live boundaries outright would risk permanently clipped turns; fixed rolling chunks (the issue's Phase 4 sketch) would cut words at arbitrary boundaries. Turn-aligned provisional rows reconciled at the end is the better design. The findings below are about specific mechanisms, not the approach.

## On the existing Greptile comments

1. **Non-transactional prune (P1)** — code is as described, but impact is overstated: prune is idempotent, retry re-runs detection→reuse→prune, and no transcript final detection wants is ever lost. The real (milder) complaint is that a transient DB error in a pure-cleanup step fails a note whose transcription fully succeeded. A single `DELETE … WHERE recording_session_id = ? AND (source, turn_index) NOT IN (…)` fixes both that and the N round trips.
2. **Start-tolerance text bleed (P2)** — much narrower than claimed. Both live and final paths coalesce same-source turns with gaps ≤ 2,500 ms, so in the common case a row starting 1,200 ms early sits entirely in detected silence. The actual loophole: coalescing only merges *adjacent* entries in the start-sorted list, so a turn from the other source interleaving between two mic turns leaves them uncoalesced with a gap as small as 901 ms (mic `merge_gap_ms` is 900) — then a live row can absorb up to ~300 ms of the previous mic turn's tail (a clipped word duplicated across two turns). System source is safe (tolerance = merge gap = 1,200). Cheap hardening: clamp the start tolerance to `min(1200, merge_gap_ms)` per source, or reject candidates overlapping the previous final turn of the same source.
3. **`sessions.insert` detaches an old loop (P2)** — unreachable in practice: session ids are fresh UUIDs, capture is single-instance, and `finish_active_capture_before_start` signal-stops any prior session. Worth the one-line defensive `if let Some(old) = sessions.insert(…) { old.stop.store(true, …) }`, severity nit.
4. **Normalized temp WAV accumulation (P2)** — confirmed, minor. Also note: `segment_path` is only removed on *success* (the `?` skips cleanup on failure), `chunks/` output is never removed per-turn, and if the process dies mid-recording the session temp dir is never re-swept (the name is session-unique and nothing cleans it at startup).

## Smaller things not worth their own thread

- During reconciliation there's a transient window where final-keyed rows (index 0..N) interleave non-chronologically with still-provisional rows keyed by `start_ms` under `ORDER BY COALESCE(turn_index, …)` — visible for seconds in the Transcription tab. Cosmetic, the PR description already acknowledges it.
- Reused rows are re-persisted with the *currently configured* provider rather than the provider that actually produced the text — metadata drift only.

Verified-clean, for the record: the `.partial.wav` chunk-walker is solid (sizes from bytes on disk, frame-aligned floor division handles torn/odd tails and stereo, unknown chunks like CoreAudio `FLLR` skipped with word-alignment, data-is-last matches both writers); the `turns.rs` refactor is a faithful extraction — line-for-line equivalent behavior vs main; provisional `start_ms` keying is collision-free per source and the index-0 collision case is upsert-resolved and tested; the kill switch works; the `record_ui_checkpoint` whitelist blocks arbitrary kinds; double provider cost is bounded at ≤2× and observable via `reusedTranscriptCount`/`jobCount` checkpoints.

## Bottom line

Keep the architecture. Before shipping I'd want: (1) a time-boxed drain with task abort, (2) an absolute uncovered-span cap on top of the 80% coverage rule, and (3) incremental RMS computation so the loop doesn't get quadratically more expensive as meetings get longer — see inline comments. I'd also resist any future temptation to loosen the tolerances or reuse live rows for note *generation* context; every loosening converts a latency optimization into a correctness risk.

🤖 Review drafted with [Claude Code](https://claude.com/claude-code)

src-tauri/src/commands.rs
junhohong: Fixed in d19b9ce: `drain` now waits at most 30s (`DRAIN_TIMEOUT`), then aborts the task and awaits the aborted handle before returning — so the loop is fully terminated before reconciliation runs, and a post-prune upsert can't happen. The aborted loop's temp dir is cleaned by the caller. `transcribe_prepared_audio` also gained an optional `stop` flag checked between chunk requests (live passes it, final passes `None`), so a stopped recording abandons remaining live chunks instead of riding out N×60s timeouts.

src-tauri/src/domain/processing.rs
junhohong: Fixed in d19b9ce: added `REUSE_MAX_UNCOVERED_MS = 2_000` — regardless of relative coverage, reused rows may leave at most 2s of the final turn uncovered, otherwise the turn is re-transcribed from the saved audio. Test: `does_not_reuse_when_the_uncovered_span_is_long_even_at_high_relative_coverage` (50s row vs 60s turn = 83% relative, rejected).

src-tauri/src/domain/processing.rs
junhohong: Fixed in d19b9ce: the join now happens during the coverage scan, and a row whose effective start lies ≥500ms behind the cursor (`REUSE_JOIN_MAX_OVERLAP_MS`) is skipped entirely — it neither counts coverage nor contributes text. If skipping it leaves too much uncovered, the absolute cap kicks in and the turn is transcribed fresh, which is the safe fallback. Test: `does_not_join_rows_that_mostly_repeat_already_covered_audio` covers the exact [0,60s]+[58s,63s] case.

src-tauri/src/audio/live.rs
junhohong: Fixed in d19b9ce: added `RmsWindowCache` — per-source windows accumulate across ticks and each tick reads only the frames appended since the previous one (only whole 30ms windows are cached; the trailing partial window is recomputed next tick, which the 4s edge guard makes irrelevant). The percentile threshold is still recomputed over all cached windows, so detection results are unchanged — `cached_detection_across_ticks_matches_detection_from_scratch` asserts parity with from-scratch detection.

src-tauri/src/commands.rs
junhohong: Fixed in d19b9ce: `finish_recording` now signal-stops the live loop when `finish_capture` fails, and the recovery discard path does `stop_and_drain` before deleting any files.

src-tauri/src/db/repositories.rs
```diff
@@ -1633,6 +1633,46 @@ impl Repositories {
...
+                continue;
+            }
+            let id: String = row.get("id");
+            sqlx::query("DELETE FROM transcripts WHERE id = ?")
+                .bind(&id)
+                .execute(&self.pool)
+                .await?;
+            pruned += 1;
+        }
+        Ok(pruned)
```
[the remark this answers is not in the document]
junhohong: Addressed in d19b9ce: the delete loop now runs inside a transaction, so pruning is all-or-nothing. (Kept the SELECT-then-delete shape rather than a dynamic NOT IN to stay within the existing query style.)

src-tauri/src/domain/processing.rs
```diff
@@ -66,15 +66,101 @@ fn source_transcript_input_from_row(row: &TranscriptDto) -> SourceTranscriptInpu
...
+        .collect::<Vec<_>>();
+    if candidates.is_empty() {
+        return None;
+    }
+    candidates.sort_by_key(|(row_start, row_end, _)| (*row_start, *row_end));
+    let mut covered = 0_i64;
+    let mut cursor = start_ms;
+    for (row_start, row_end, _) in &candidates {
+        let overlap_start = (*row_start).max(cursor);
+        let overlap_end = (*row_end).min(end_ms);
```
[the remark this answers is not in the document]
junhohong: Addressed in d19b9ce, in the narrower form that's actually reachable: both tolerances are now clamped to the source's detection merge gap (`min(tolerance, merge_gap_ms)`), so a row can never start/end inside a neighboring same-source turn — including the interleaved-source case where two mic turns end up only ~901ms apart. Test: `start_tolerance_is_clamped_to_the_source_merge_gap`.

src-tauri/src/domain/live_transcription.rs
```diff
@@ -0,0 +1,722 @@
...
+            note_id,
+            task_session_id,
+            source_mode,
+            sources,
+            task_stop,
+        )
+        .await;
+    });
+    let mut sessions = LIVE_SESSIONS
+        .lock()
```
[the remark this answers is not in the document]
junhohong: Addressed in d19b9ce: `sessions.insert` now signal-stops a displaced handle (unreachable via current code paths, but the invariant is explicit).

src-tauri/src/domain/live_transcription.rs
```diff
@@ -0,0 +1,722 @@
...
+                    turn.end_ms,
+                    provisional_index,
+                )
+                .await
+            {
+                Ok(row) => {
+                    tracing::info!(
+                        session_id = %context.session_id,
+                        source = %turn.source,
+                        start_ms = turn.start_ms,
```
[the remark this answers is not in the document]
junhohong: Addressed in d19b9ce: each live turn's segment, normalized, and split-chunk WAVs are removed after the attempt, on success and failure alike.

PR #131, opened 2026-06-10
Floating meeting-recording HUD + frosted dictation pill

andrewwashuta:
## What

A floating **meeting-recording pill** — a small always-on-top indicator that appears when the main window is closed or minimized while a meeting is recording — plus a restyle of the dictation HUD so both pills share one visual language.

### Meeting recording HUD (new)
- Frosted always-on-top pill (record-LED breath + live 4-bar waveform) that appears only when the main window can't show the in-app recorder bar, driven by a Rust supervisor reading capture status directly
- Pure presence indicator: clicking it reopens the app on the **meetings view** for the recording's note (covered by a regression test); all recording controls stay in-app
- Real behind-window blur: the surface is a native `NSVisualEffectView` sized to the pill behind the webview (CSS `backdrop-filter` can't sample other apps' pixels), depth from the native `NSWindow` shadow
- Orientation-aware: parked in a left/right screen third it stands upright via one native Core Animation quarter-turn on the contentView (frost + tint + DOM turn as a unit; the bars counter-rotate in CSS so the waveform keeps reading left-to-right); drag-settle + mouse-button guards prevent mid-drag flips
- Non-activating `NSPanel` — clicking never steals focus from the meeting app

### Dictation HUD (consistency restyle)
- Same frosted vibrancy surface, darker 72% popover tint, and native shadow as the meeting pill; window sized exactly to the pill per state
- State-width changes morph: contents crossfade while the glass eases to the new frame; native alpha drives the exit dissolve; the error wobble is a native window shake
- Deliberately **stays flat**: a side-zone upright turn for the dictation pill was built and then reverted in-branch — it read as acrobatics, not utility (Wispr keeps theirs fixed in place). The quarter-turn stays a meeting-pill-only trick. History retained for future reference.

### Shared plumbing
- `src/lib/recorder-levels.ts`: waveform shaping/source-mixing helpers extracted from the React `Waveform` component so the vanilla-TS meeting HUD shares one source of truth (re-exported for existing importers)
- `capture::current_status()` exposes the live recording snapshot to the supervisor

## Testing
- `cargo check` + `cargo test` (103 passing, includes the dictation shortcut/wake-word suites)
- `npx vitest run src/test` — 213 passing, including new meeting-HUD webview tests and the reopen-flow regression test
- Manual: pill appears on close/minimize mid-recording, hides on restore/stop; upright turn in side thirds; click-to-reopen lands on the meeting note; dictation states (listening → transcribing → pasting → errors) morph cleanly


andrewwashuta:
@greptile-apps

PR #175, opened 2026-06-10
Warm intro-screen fade + scoped chat-bar handoff animation

andrewwashuta:
## What

Two pieces of new-session polish:

**Warm fade on the intro screen.** The hero now shows a subtle terracotta wash (`--brand` at 11%) welling up from the bottom of the card, edge to edge:
- The workspace gets its own stacking context (`isolation: isolate`) so the `z-index: -1` gradient paints over the card instead of vanishing behind it.
- The wash bleeds past `.main-panel-body`'s 24px gutters to reach the card's edges and rounded corners.
- The card's 96px bottom dissolve overlay (`.main-panel::after`) is hidden while the hero is up — its scroll timeline is inactive there (no `.agent-scroll`) and it was bleaching the wash back to white.
- Fades out on opacity as the composer docks; disabled under reduced motion.
- Suggestion-chip icons pick up the same terracotta to echo the wash.

**Chat bar animates only on real thread starts.** The hero→dock composer FLIP fired on any hero exit, including switching to an existing chat from the sidebar. It's now armed only when `submitHermesSession` creates a brand-new session, and the flag is cleared while the hero is up so a failed submit can't leak a stale glide. The composer auto-grow effect also skips its height glide across hero transitions — it was double-animating the box against the FLIP on handoff (visible jitter).

| Action | Before | After |
|---|---|---|
| Switch to an existing chat from the hero | bar glided + height-animated | instant swap |
| Empty chat → start a thread | glide + competing height animation | single clean glide |

## Validation

- `tsc --noEmit` clean; all 32 agent-workspace tests pass.
- Verified visually in `tauri:dev`: fade fills the card to the bottom corners, dissolve overlay returns with the conversation scroller, sidebar switches are instant, thread-start glide is smooth.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -675,6 +675,11 @@ export function AgentWorkspace({
   const liveEventsRef = useRef<Record<string, LiveHermesEvent[]>>({});
   const hydratedTaskIdsRef = useRef<Set<string>>(new Set());
   const newSessionModeRef = useRef(false);
+  // True only while a brand-new thread is being started from the hero. The
+  // hero→dock composer FLIP keys off this so it glides *only* when the empty
+  // chat hands over to a fresh thread — not when the hero is dismissed by
+  // selecting an existing chat from the sidebar (that should swap instantly).
+  const heroExitViaThreadRef = useRef(false);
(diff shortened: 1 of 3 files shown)


andrewwashuta:
@greptile-apps

PR #187, opened 2026-06-11
Add dev flag to replay onboarding and force logout

seriusanbudi:
## Summary
- Add `pnpm tauri:dev --replay-onboarding` as a dev-only replay path.
- Clear onboarding completion and resume state, then force OS Accounts logout before the account snapshot loads.
- Document the workflow in the README.

## Testing
- Added unit tests for onboarding replay reset behavior.
- Added a unit test covering forced logout before account status loads.
- Full Vitest suite passed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,6 +25,13 @@ pnpm install
 pnpm tauri:dev
 ```
 
+To replay first-run onboarding, clear the saved onboarding checkpoint, and log
+out of OS Accounts without wiping the rest of your local app data:
+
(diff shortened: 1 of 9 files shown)


seriusanbudi:
Verified the review issue. If the forced logout rejects during replay startup, the current effect skips the account status refresh. I’m going to handle logout failure locally and still load the current account snapshot so the UI does not silently settle on the empty fallback.

PR #191, opened 2026-06-11
Let plan credits satisfy the membership gate

0xgaut:
## Problem

Users on credits-based plans (prepaid, top-ups, promos) get pitched the free trial during onboarding and end up stranded on **"Waiting for your trial to start"** — there's no trial to start, and no way past the step. The post-onboarding `TrialGate` blocks them the same way.

Cause: every membership check (`shouldBlockOnTrial`, the onboarding trial step's skip, its back-navigation guard) only looks at `subscription.status`. A credits account has no Stripe subscription, but the backend happily `authorize`→`charge`s it — so the gate was asking these users to pay twice.

## Fix

- New `hasPlanCredits(account)` in `account-gate.ts`: positive `balance.credits` or `balance.usdMillis`.
- `shouldBlockOnTrial` returns false for credit holders (zero-balance non-subscribers still blocked; unknown subscription state with no credits still blocked).
- Onboarding `TrialStep` and the flow's back-nav guard skip on the same condition.

## ⚠️ Policy change — flagging for review

`#181` explicitly decided *"credits alone do NOT grant access"* (cancelled subscribers with leftover top-ups land on the gate). This PR deliberately reverses that for accounts holding a balance: they get access until the balance runs dry, then hit the gate as before. The two tests pinning the old policy were flipped to match, plus a new test that a spent balance still blocks.

## Tests

- `account-gate.test.ts`: credits allow access, spent balance blocks, `hasPlanCredits` unit coverage.
- `onboarding.test.tsx`: a credits account walks Welcome → Privacy → Permissions → Set up straight into dictation practice — no trial pitch, no checkout call.
- Full suite: 313 passing.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -17,7 +17,6 @@
         "central-icons": "npm:@central-icons-react/round-outlined-radius-3-stroke-2@1.1.233",
         "central-icons-filled": "npm:@central-icons-react/round-filled-radius-3-stroke-2@1.1.241",
         "framer-motion": "^12.39.0",
-        "lucide-react": "^0.468.0",
         "react": "^18.3.1",
         "react-dom": "^18.3.1",
         "unicode-animations": "^1.0.3",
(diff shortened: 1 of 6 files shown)


0xgaut:
Closing: product decision is that membership requires a subscription — credits alone shouldn't grant access (the #181 policy stands). The real bug is the portal's start-trial endpoint failing with a non-JSON error; tracking that separately.

PR #201, opened 2026-06-11
Scope Unrestricted to the session that opted in

junhohong:
## The report

> When I start a session in Unrestricted mode, the Unrestricted pill appears on EVERY session. Is that how it works? I toggle it, and it changes it for every agent session?

**Yes — that was exactly how it worked, and the pill was honestly reporting it.** The Seatbelt write-jail is applied when June's runtime process spawns, and one process serves all sessions. Starting one Unrestricted session restarted the runtime unsandboxed; from then on, **a follow-up sent to any other session — including ones created as Sandboxed — silently ran with full write access**, because follow-ups reused whatever runtime was up (`ensureHermesGateway(undefined)`). The badge keyed off the live runtime, so it appeared everywhere. The badge wasn't the bug; the behavior was.

## The fix

The opt-in is now genuinely per session:

- **Each session records its mode at creation** (`june.agent.unrestrictedSessions` in localStorage — absence means sandboxed, so every pre-existing session falls back to the safe default).
- **Every send enforces the target session's mode.** Following up on a sandboxed session while the runtime is unrestricted restarts the runtime back into the jail before the prompt is submitted; following up on the opted-in session restores Unrestricted if the runtime has since dropped back. This reuses the exact restart path new-session creation already had — the sandbox still can't change on a live process, so mode changes mean a runtime restart.
- **The badge describes the selected session, not the live process.** A sandboxed session shows no pill even while an Unrestricted runtime from another session is still up — truthful now, because that session's own sends re-apply the jail. Hover copy updated to explain this.
- The record is dropped when a session is deleted.

## Known trade-off

Sessions of different modes can't run concurrently (one process, one jail). Switching modes mid-flight kills the other session's in-flight run — same trade-off the new-session restart already made; the existing working-state reconciliation handles it. If concurrent mixed-mode sessions ever matter, that's a per-session-process architecture change, noted here deliberately rather than smuggled in.

## Tests

- Badge: shows only on opted-in sessions; stays off a sandboxed session even while the runtime is unsandboxed.
- Enforcement: a follow-up to a never-opted-in session restarts the bridge sandboxed (`startHermesBridge(undefined, false)`) before `prompt.submit`; a follow-up to an opted-in session restores Unrestricted after the runtime dropped back.
- 332 frontend tests, tsc, prettier pass.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -142,6 +142,11 @@ import {
 } from "../../lib/model-privacy";
 import { messageFromError } from "../../lib/errors";
 import {
+  forgetSessionMode,
+  rememberSessionMode,
+  sessionUnrestricted,
+} from "../../lib/agent-session-modes";
+import {
   buildAgentChatTurns,
   buildHermesSessionChatTurns,
   type AgentApprovalChoice,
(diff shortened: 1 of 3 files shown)


junhohong:
Folded into #207 (now based on main) so the per-session scoping and the side-by-side runtime pair land as one reviewable piece — including the Greptile catch here (the sidebar delete path now drops the session's Unrestricted record; fixed in 2d7eaab).

PR #213, opened 2026-06-11
Route every verify link through Rust to the scribe-api /verify page

junhohong:
## What was broken

The onboarding privacy footnote ("Verify it yourself: how routing, retention, and attestation work.") pointed at the generic `opensoftware.network/privacy` page rather than the API's `/verify` attestation walkthrough. Worse, it did nothing when clicked - and so did the other two verify surfaces.

**Why every one of these was a dead button:** the app never installs a new-window handler (`on_new_window`), and wry's WKWebView UI delegate returns `nil` from `createWebViewWithConfiguration` in that case - `target="_blank"` navigations are silently dropped (verified in the vendored wry 0.55.1 source). It's the same reason the accounts-portal links were moved behind a Rust command back in 945d1bd. The settings "Verify server" button and the agent privacy badge from #206 shipped as anchors and were equally inert.

## The fix

One command, three consumers:

- **Rust:** `scribe_verify_url` (returned a URL for anchors that can't navigate) becomes `scribe_open_verify_page`, which opens `{SCRIBE_API_URL}/verify` in the default browser through the same `open_in_browser` helper the portal links use.
- **Onboarding footnote:** now a link-styled button invoking the command - lands on the API's /verify page as requested.
- **Settings -> About "Verify server":** anchor -> button, URL-fetch state removed.
- **Agent privacy badge:** anchor -> button; the "no verify URL -> plain tooltip" branch is gone because the URL always resolves (it only guarded the fetch latency window), so the badge always click-throughs to the proof.

## Testing

- 337/337 tests pass; `tsc` and `cargo check`/`fmt` clean.
- Updated the three surface tests to assert the command is invoked on click (instead of asserting `href`/`target` attributes on anchors that never worked), plus a new onboarding test for the footnote.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -546,9 +546,13 @@ pub async fn check_recording_source_readiness(
         .map_err(|error| AppError::new("readiness_check_failed", error.to_string()))
 }
 
+/// Opens the scribe-api `/verify` page (enclave attestation, routing,
+/// retention) in the default browser. Must route through Rust: the webview
+/// installs no new-window handler, so `target="_blank"` anchors are silently
+/// dropped — same reason the accounts portal links go through a command.
 #[tauri::command]
-pub fn scribe_verify_url() -> String {
-    crate::scribe_api::verify_url()
+pub fn scribe_open_verify_page() -> Result<(), AppError> {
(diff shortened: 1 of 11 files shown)


junhohong:
@greptile review

PR #218, opened 2026-06-11
[codex] Stabilize agent thinking disclosure

andrewwashuta:
Fixes the agent thinking disclosure so user-opened live thinking stays open while reasoning continues streaming.

The active thinking row now stores open state at the workspace level, and the initial Thinking placeholder uses the same text sizing as the full disclosure to avoid a small-to-large flash.

Adds a regression test for keeping Thinking open across additional reasoning deltas.

Validated with `pnpm exec tsc --noEmit`, `pnpm exec vitest run src/test/agent-workspace.test.tsx`, and `git diff --check`.


andrewwashuta:
@greptile-apps

PR #225, opened 2026-06-11
File issue reports as Bug-tagged issues in os-platform

junhohong:
## Why

In-app bug reports tell the user "this is being sent to the June team", but server-side they only went to a generic webhook — or, with none configured, the structured logs. This makes the pipeline real: every report becomes an **Issue in the os-platform tracker, tagged `bug`**, with the agent's diagnosis and the report metadata attached.

**Uses only os-platform's stock API — no platform-side changes required** (the earlier create-time `label_slugs` companion, os-platform#72, was deliberately dropped).

## What

**New `OsPlatformIssueReportSink`** (implements the existing `IssueReportSink` port):

1. **Attachments upload first** to `POST /v1/files` — best-effort: an unreadable/oversized/rejected upload never blocks the report, and the user's attachment names are listed in the body either way.
2. **Issue created** at `POST /v1/orgs/{org}/projects/{project}/bounties` with `type: bug`, `status: todo`, `reward_amount_units: "0"`, and the uploaded `file_ids`. Title is the report's first line (`June report: …`, char-boundary truncated); body carries the description, the agent diagnosis, and a metadata section (reporter, session, app version, platform).
3. **Bug tag attached second** via the existing `PUT /v1/orgs/{org}/bounties/{number}/labels` (set-replace is safe on a just-created Issue — nothing to clobber). First report into a Project: the label doesn't exist yet, so the missing-label rejection creates it (`Bug`, `#ef4444`) and retries once. Tagging is **best-effort after the Issue exists** — a permanently denied label logs a warning but never fails the delivery, because failing would re-show the report to the user as unsent while the Issue already sits in the tracker (and it carries `type: bug` regardless).

**Sink selection** in `main.rs`: os-platform when configured → webhook → log-only, first match wins.

**Config** (`[issue_reports]`, documented in config.toml): `os_platform_api_url`, `os_platform_api_key` (fellow `osk_…` key, env-only, redacted Debug), `os_platform_org`, `os_platform_project`, `os_platform_label` (default `bug`). All default-empty, so existing deployments are untouched until configured.

## Ops notes

The `osk_` key belongs to a **bot user**: create a dedicated OS Accounts account, sign it into os-platform once, invite it to the target Org (member role suffices), then as the bot `POST /v1/me/api-keys` — the plaintext key is returned exactly once.

## Tests

wiremock coverage: happy path (attachment uploaded, issue created, labels PUT asserted on the wire), first-use label creation + PUT retry, label-attachment-denied → delivery still succeeds, create rejection → upstream error, config gating. Full scribe-api suite (100 tests), clippy, fmt clean.

### DIFF
--- a/scribe-api/config.toml
+++ b/scribe-api/config.toml
@@ -13,10 +13,19 @@ image_repo = "ghcr.io/open-software-network/scribe-api"
 trust_center_url = "https://trust.phala.com/app/15f8d2fd586da8b99c6082b3c2cba64127ceeb8c"
 
 [issue_reports]
-# Webhook that receives user-submitted issue reports as a JSON POST. Often
-# embeds a secret token, so set it per environment via
-# SCRIBE__ISSUE_REPORTS__WEBHOOK_URL. Empty keeps reports in the structured
-# logs only.
+# Where user-submitted issue reports go, first match wins:
+# 1. os-platform issue tracker — reports become Issues tagged with
+#    os_platform_label in the configured Org/Project. The API key is a
+#    fellow key (osk_...) of a bot user that is a member of that Org;
+#    set it per environment via SCRIBE__ISSUE_REPORTS__OS_PLATFORM_API_KEY.
+# 2. Webhook — each report as a JSON POST. Often embeds a secret token, so
+#    set it per environment via SCRIBE__ISSUE_REPORTS__WEBHOOK_URL.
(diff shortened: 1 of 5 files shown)


junhohong:
Parking alongside open-software-network/os-platform#72 — the feature isn't ready yet. Branch (feature/os-platform-issue-sink) stays intact; reopen together with the os-platform half.

PR #227, opened 2026-06-11
Copy truth audit: align claims with what the product actually does

junhohong:
Audited every user-facing claim in the app (onboarding steps, account/trial gates, settings, agent UI, SOUL.md persona, Info.plist permission strings) against the implementation, and fixed the ones that are not true today.

## Fixed - was false or contradicted by the app itself

1. **"Our code is open source"** (privacy card; SOUL.md says the backend is open source). The repo is **private** (verified via the GitHub API). What IS true: the Hermes agent framework is public, and the backend runs in an attested TEE you can inspect via /verify. The copy now says exactly that, keeping the "You don't have to trust us. You can check." punch - which the attestation genuinely backs.
2. **"Nothing stored, nothing trained on. Ever."** and **"That's always on."** The app's own model picker offers anonymous-mode models whose label says "the model provider may retain prompts." Both claims now scope to the default and point at the labels. (Checked that the new GLM 5.1 default is a Venice private/zero-retention model, so "by default" holds.)
3. **"The agent asks before it edits or deletes files, sends anything, or spends anything."** Reality: hermes's approval gate is a dangerous-command gate, the sandbox blocks writes outside the workspace, and agent turns meter credits without asking (every chat settles a charge). Rewritten around the real controls: sandbox by default, approval for risky commands, full activity log, stop anytime.
4. **"who said what"** (two places). The turn pipeline separates your mic from system audio - explicitly no speaker diarization per the feature spec. Now "your side and theirs".
5. **Mic permission copy** said June listens only while dictating or recording; the settings mic test also listens. Now "only when you ask June to listen".
6. **"inference runs through zero-retention models"** in the data-practices card gains the "by default" qualifier the sibling copy already carries.

## Checked and deliberately left alone

- "Dictate into any app" / "type anywhere": conventional puffery; accessibility paste works in effectively all apps.
- "Files, sessions, and memory... never mirrored to a cloud": true (hermes home is local).
- "What we store: account, login, and billing records": true for ambient collection; issue reports are explicit user submissions with consent visible in the flow.
- "Backend runs in a secure enclave": true (Phala CVM, attested).
- Trial copy ("cancel anytime", "$0 due today", end-date timeline): consistent with the subscription flow.
- E2EE/private/anonymous badge descriptions: already accurate and honest.

## Flags for follow-up (not in this PR)

- `FALLBACK_TRIAL_LENGTH_DAYS = 14` must match the live Stripe price's trial_period_days - not verifiable from code; worth a glance at the Stripe dashboard.
- README claims dictation "does not... store transcript records," but dictation history retains transcripts locally for 7 days. README is dev-facing so it's out of scope here, but it's stale.

368/368 tests pass, lint and cargo check clean. No tests pinned the changed strings.

### DIFF
--- a/src-tauri/src/hermes_bridge.rs
+++ b/src-tauri/src/hermes_bridge.rs
@@ -54,7 +54,7 @@ Privacy is your defining trait, by architecture rather than promise. When asked 
 
 - You run locally on the user's desktop. Files, sessions, memory, and agent state stay on the user's disk by default.
 - Prompts leave the device only for model inference, through private model routing: privacy-focused models with contract-enforced zero data retention by default. If the user opts into third-party models, identifying metadata is stripped first.
-- June's backend runs in a TEE with cryptographic attestation, so users can verify what is running rather than trust it. The agent framework you run on (Hermes) is open source. The service stores only account, login, and billing records.
+- June's backend is open source and runs in a TEE with cryptographic attestation, so users can verify it rather than trust it. The service stores only account, login, and billing records.
 - Open Software never trains on the user's data.
 
 You are helpful, knowledgeable, and direct. Communicate clearly, admit uncertainty when appropriate, and prioritize being genuinely useful over being verbose. Be targeted and efficient in your exploration and investigations. Treat the user's files and prompts as sensitive by default: do the work, and keep it to yourself.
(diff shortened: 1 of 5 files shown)


junhohong:
Updated after the license change: the repo is Apache 2.0 (#224) and goes public at launch, so the two open-source claims (privacy card + SOUL.md) are restored to their original wording - they're true for everyone who will read them. The other five fixes (retention scoping, approvals, diarization, mic copy, default qualifier) stand. Branch rebased on main; 381/381 tests pass.

PR #231, opened 2026-06-11
Fix approval explanations arriving cut off mid-word

junhohong:
## The bug (from Gaut's Slack screenshot)

"Explain first" on an approval card rendered: *"This checks whether your ANTHROPI"* - cut mid-word.

**Root cause:** `explain_agent_approval` capped the side call at `max_tokens: 220`. That was fine until #220 made GLM 5.1 the default text model - a reasoning model whose hidden thinking spends from the same token budget as the visible answer. By the time it finished thinking, a handful of tokens remained for the actual explanation, and generation stopped mid-word (`finish_reason: "length"`). The session-title side call (`max_tokens: 24`) has the same exposure, just less visibly.

## Fix, three layers

1. **Budgets sized for reasoning models:** explanation 220 -> 2000, title 24 -> 500. The visible outputs stay short (the prompts demand it); the headroom is for thinking.
2. **`<think>` stripping:** `extract_chat_completion_text` now removes inline `<think>...</think>` spans some providers return in `content`, and treats an unterminated block (cap landed inside the reasoning itself) as "no answer".
3. **No mid-word fragments, ever:** when `finish_reason` is `"length"`, the text is cut back to its last complete sentence. A fragment with no complete sentence becomes a generation error (the card's existing error state) instead of garbled copy.

## Testing

Four new unit tests covering the exact screenshot case (capped fragment -> error, capped multi-sentence -> trimmed to last sentence), think-block stripping, and cap-inside-reasoning. `cargo test scribe_api` 8/8, clippy clean, fmt clean.

### DIFF
--- a/src-tauri/src/scribe_api.rs
+++ b/src-tauri/src/scribe_api.rs
@@ -505,7 +505,10 @@ pub async fn suggest_agent_session_title(prompt: &str) -> Result<String, AppErro
             }
         ],
         "temperature": 0.1,
-        "max_tokens": 24
+        // Sized for reasoning models: hidden thinking spends from the same
+        // budget as the 2-to-5-word title, and a cap hit mid-think yields no
+        // title at all.
+        "max_tokens": 500
     }))
     .await?;
     if !(200..300).contains(&response.status) {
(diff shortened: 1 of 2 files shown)


junhohong:
Bumped per review: explanation budget is now 8000 tokens, the prompt scales length to the request instead of hard-capping at 4 sentences (short for simple commands, short paragraphs when a complex script needs them), and the card renders blank-line breaks as real paragraphs (it previously collapsed everything into a single <p>). 381/381 tests, lint and fmt clean.

PR #234, opened 2026-06-11
Show the current model in the session bar and switch models from it

junhohong:
Implements Gaut's feature request: the privacy pill in the agent session bar ("Anonymous mode" in the screenshot) should lead to changing models when clicked, and the current model should be visible top right.

## What changed

**The pill is now the model control.** Top right of the agent workspace it reads `[privacy icon] GLM 5.1 · Private mode`: the current text model by name, with its privacy mode trailing as secondary text. When the bar runs tight (the existing container query), it sheds the privacy label and keeps the model name. A model that is configured but missing from the catalog still shows as a name-only pill rather than disappearing.

**Clicking it opens the model picker in place.** The same picker dialog as Settings -> Models (Suggested / All tabs, search, tool-capability guard), opened right in the workspace instead of bouncing the user to Settings. Selecting a model writes the global text-model setting and broadcasts the existing `provider-model-settings-changed` event, so the pill, the task-detail header, and Settings all refresh through the one path; a failed switch lands in the workspace error banner.

**The attestation walkthrough is not lost.** The pill used to open the server verification page; that page stays reachable from Settings and onboarding, where the claim is explained in context. The pill's tooltip still carries the privacy mode description, now ending in "Click to change the model."

**Refactor:** `ModelPickerDialog` and its catalog helpers (`ModelMeta`, `selectedModel`, `modelOptions`, pricing/context labels) moved verbatim from `AppSettings.tsx` into `src/components/settings/ModelPickerDialog.tsx` so the workspace can open the picker without importing the whole settings surface. `AppSettings.tsx` shrinks by ~370 lines and keeps identical behavior.

**Drive-by:** syncs `bun.lock` with the `border-beam`/`agentation` dependencies that landed in `package.json` with the onboarding redesign (#229) but missed the bun lockfile (CI uses pnpm, so this only affects local bun installs).

## Testing

- Updated the pill tests: accessible name now leads with the model name; the verify-page click test became "opens the model picker from the model pill".
- New end-to-end switch test: open picker -> All tab -> select a different model -> asserts `setVeniceModel("generation", ...)` is called and the pill re-renders with the new model name and privacy mode.
- Settings tests unchanged and passing against the extracted dialog.
- `tsc` clean, `vite build` clean, full suite passes (40 files, 392 tests).


andrewwashuta:
@greptileai

andrewwashuta:
@greptileai

PR #277, opened 2026-06-12
Defer the system audio permission prompt to the first record action

junhohong:
## Problem

> when a user first opens June after the onboarding, they get hit with "June would like to record system audio" permission out of the blue

The post-onboarding mount effect in `App.tsx` called `check_recording_source_readiness("microphonePlusSystem")`, and that command unconditionally launched the system-audio capture helper to verify the TCC grant. Launching the helper is what fires the native permission prompt, so fresh installs saw it at app open instead of when they actually tried to record. The window-focus refresh had the same side effect.

## Fix

`check_recording_source_readiness` now takes a `probeSystemPermission` flag (default `false`, serde-defaulted so older callers are unaffected):

- **Passive** (startup mount, focus refresh): reads device and macOS-version readiness only; system permission state stays `unknown` and no helper launches, so no prompt.
- **Probe** (`handleStartRecording` in the frontend, `start_recording` preflight in Rust): launches the helper as before, surfacing the TCC prompt at the moment the user starts a recording.

One nuance kept intact: after an observed denial, the focus refresh still probes so that a user who flips the toggle in System Settings and returns to June gets the system-audio toggle back without recording first. That path never shows a dialog, since macOS does not re-prompt a denied TCC entry.

Behavior summary:
- Fresh install: no prompt at launch. First click of record asks for system audio in context; deny falls back to mic-only for that take (existing fallback), grant records everything.
- Previously denied: the blocked strip appears after the first record attempt, and granting in System Settings is picked up on refocus as before.

## Testing

- `cargo test --lib`: 135 passed, including new gate test `skips_system_audio_permission_probe_unless_requested`
- `bun run test`: 428 passed; contract test now pins both the passive and probing request shapes
- `tsc --noEmit` clean; clippy warning count unchanged from main

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -550,9 +550,11 @@ pub async fn check_recording_source_readiness(
     // The system-audio permission probe can block for over a minute while the
     // helper waits on a CoreAudio permission grant; keep that work off the
     // async runtime so other commands stay responsive.
-    tokio::task::spawn_blocking(move || recording_source_readiness(request.source_mode))
-        .await
-        .map_err(|error| AppError::new("readiness_check_failed", error.to_string()))
+    tokio::task::spawn_blocking(move || {
+        recording_source_readiness(request.source_mode, request.probe_system_permission)
+    })
+    .await
+    .map_err(|error| AppError::new("readiness_check_failed", error.to_string()))
 }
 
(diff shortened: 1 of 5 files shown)


andrewwashuta:
https://github.com/open-software-network/os-june/pull/272 @junhohong was this already solved by the earlier story?

PR #301, opened 2026-06-15
Browser-style tabs bar

andrewwashuta:
## Summary

Adds a Linear/Arc-style document tabs feature to the main nav bar, letting you browse multiple notes/sessions/agents simultaneously without losing navigation state.

- **Tabs = snapshots**: each tab stores a complete navigation state (active view, selected item, breadcrumb origin)
- **Live capture**: every nav change auto-syncs to the active tab, so existing handlers need zero changes
- **Smart restore**: switching tabs replays their snapshot with safeguards against async note-load races
- **Responsive & dense**: tabs start at ~¼ width, shrink to icons, then overflow to a popover when needed
- **Full interactions**: ⌘T (new), ⌘[/] (cycle), ⌘1-9 (jump), right-click (close/close-others), middle-click (close), hover tooltip

## Implementation

- `TabNav` model captures `view + selected entity + breadcrumb`, with equality checks for cheap diffs
- Capture effect mirrors live nav into active tab on every change
- Apply effect restores a snapshot with a guard that releases once async work settles
- Tab strip layout is a flex container that shrinks/scrolls responsively
- Overflow popover (Figma-style) lists all off-screen tabs; active tab always stays visible
- `HoverTip` extended with optional `width` and `delay` for the compact `+` button tooltip

## Files

- `src/app/tabs/tabs.ts` — `TabNav` model, equality, defaults
- `src/components/tabs/TabBar.tsx` — the pill strip, overflow popover, responsive sizing
- `src/test/tabs.test.ts` — snapshot equality + state transitions
- `src/app/App.tsx` — tab engine (capture/restore/operations), handlers wired
- `src/styles/app.css` — layout restructure, tab styling, responsive collapse
- `src/styles/tokens.css` — new `--card-top` pin
- `src/components/ui/HoverTip.tsx` — new `compact` variant and `delay` prop

## Testing

All 484 tests pass. New tab-model tests lock in capture/restore semantics.

Verified:
- Typecheck ✅
- Build ✅
- Test suite ✅

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -11,7 +11,11 @@ import {
   useRef,
   useState,
 } from "react";
-import type { CSSProperties, PointerEvent as ReactPointerEvent } from "react";
+import type {
+  CSSProperties,
+  PointerEvent as ReactPointerEvent,
+  ReactNode,
+} from "react";
 import { AccountGate, JuneMark } from "../components/account/AccountGate";
 import { TrialGate } from "../components/account/TrialGate";
 import { OnboardingFlow } from "../components/onboarding/OnboardingFlow";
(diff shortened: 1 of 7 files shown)


andrewwashuta:
@greptileai I've pushed changes addressing your review.

**Addressed**
- **`folderReturnTarget` not cleared on tab restore (P1)** → `applyNav` now clears `setFolderReturnTarget(undefined)` on every restore, alongside the other per-view origin resets, so the "back to <note>" breadcrumb can't leak across tabs.
- **`settingsReturnView` not populated when restoring a settings tab (P2)** → `applyNav` now records the return view for settings tabs (mirroring `openSettings`), so exiting Settings lands back where the tab came from.
- **Overflow button label vs popover content (P2)** → the popover is intentionally a full tab switcher (it keeps every tab readable once the strip collapses to icons), so the `aria-label` now reads "Show all N tabs" to match the content. The numeric badge still indicates how many tabs are currently off-strip.

All 511 tests pass; typecheck + build clean.

Please take another look.

PR #311, opened 2026-06-16
Prevent source labels leaking into generated notes

andrewwashuta:
## Summary
- Mark transcript source labels as metadata in note generation prompts
- Strip leaked leading `Microphone:` / `System:` prefixes from generated note lines as a deterministic guard
- Bump the note prompt version to `notes-mvp-v4`

## Verification
- `cargo test --manifest-path scribe-api/Cargo.toml -p scribe-providers venice::tests --locked`
- `cargo test --manifest-path scribe-api/Cargo.toml -p scribe-providers --locked`
- `cargo check --manifest-path src-tauri/Cargo.toml --locked`


andrewwashuta:
@greptileai I've pushed changes addressing the review/CI state.

**Addressed**
- CI clippy failure: changed `map(...).unwrap_or_else(...)` to `map_or_else(...)` in `strip_generated_source_label`.

**Skipped (with reason)**
- (Greptile) Conditionally emit `<transcript_source_metadata>` only when labels are present: keeping this always-on as a small, explicit safety instruction for generation. Greptile marked this as defensible and safe to merge.
- (Greptile) Remove the final trim in `cleanup_generated_note_text`: keeping it as defensive cleanup because the helper is tested directly and should stay robust if `first_choice_text()` changes later.

Please take another look.

andrewwashuta:
@greptileai I've pushed one more follow-up after reviewing the Codex advisory comments.

**Addressed**
- (Codex) Source-label metadata/cleanup should not apply to single-source transcripts → added an explicit `transcriptSourceLabels` request flag, set false for single-source generation and true for dual-source labeled transcript generation. Runtime source-label prompt instructions and output cleanup now only run when that flag is true.
- (Codex) Markdown lines could still leak source labels behind markers → source-label cleanup now handles common Markdown prefixes such as bullets, blockquotes, headings, and ordered lists.
- Added regressions for preserving spoken `System:` text in single-source notes, omitting source-label metadata for unlabeled transcripts, and stripping labels after Markdown markers.

**Skipped (with reason)**
- None.

Please take another look.

andrewwashuta:
@greptileai I've pushed one final follow-up for the latest Codex advisory.

**Addressed**
- (Codex) Preserve spoken source-like prefixes in dual-source transcripts → made cleanup transcript-aware. It now strips a generated source prefix only when the stripped text matches the spoken content after removing the transcript's outer source label. This preserves a correct line like `System: restart the service` from `Microphone: System: restart the service`, while still cleaning a leaked outer label like `Microphone: System: restart the service`.
- Added regression coverage for this nested source-like prefix case.

**Verification**
- `cargo test --manifest-path scribe-api/Cargo.toml -p scribe-providers --locked`
- `cargo clippy --manifest-path scribe-api/Cargo.toml --all-targets --all-features --locked -- -D warnings`

Please take another look.

PR #320, opened 2026-06-16
[codex] Align note tab title and chrome

andrewwashuta:
## Summary
- Make empty note tabs use the same `New note` label as the editor and note lists.
- Align the first window chrome tab with the main panel edge without clipping its rounded corner or shadow.
- Add regression coverage for the blank-note tab label.

## Validation
- `pnpm test -- src/test/app-shortcut.test.tsx`
- `pnpm run lint`
- `pnpm run build`

## Notes
- Screenshot QA could not be automated in this environment because Browser control and Playwright/Chrome were unavailable.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -206,7 +206,7 @@ function tabMeta(
         ? notes.find((n) => n.id === nav.noteId)
         : undefined;
       return {
-        title: note?.title?.trim() || "Untitled note",
+        title: note?.title?.trim() || "New note",
         icon: <IconNoteText size={TAB_ICON_SIZE} />,
       };
     }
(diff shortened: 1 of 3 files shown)


andrewwashuta:

**Addressed**
- Greptile: `querySelector` returning `null` produces an opaque test failure -> switched the regression assertion to Testing Library `screen.getByRole("tab", { name: "New note" })` and assert the active tab attribute.
- Greptile: `overflow: visible` removes transient-resize clipping -> replaced unrestricted visible overflow with `overflow: clip` plus `overflow-clip-margin` so resize spillover stays clipped while the flush first tab keeps its corner and shadow breathing room.

Please take another look.

PR #310, opened 2026-06-16
Add in-app referral link screen

0xgaut:
## Summary
- add a sidebar Invite a friend entry above the account menu
- show an in-app referral dialog with link copy and earned-month stats
- fetch the referral link through the Rust OS Accounts bridge so users do not leave June to get it

## Verification
- pnpm lint
- pnpm test
- pnpm test:rust

PR #323, opened 2026-06-16
Rework referral entry point and dialog

andrewwashuta:
## What

Cleanup pass over the in-app referral work (#310), driven by Mobbin references.

### Entry point
Moves **Invite friends** out of the sidebar footer (it sat as a standalone nav item with a permanently greyed icon when signed out) into the **account menu** as the top row, with a terracotta gift icon so it pops. This matches how invite/referral entry points consistently live inside the account/profile menu across the apps surveyed (Gopuff, PayPal, Vivid, Cash App).

### Dialog redesign
Rebuilds the referral dialog as a **50/50 split** (Heidi / Base44 / Lovable style):
- **Left:** a terracotta hero carrying the offer, built from our own kit — the fixed `--brand` gradient (theme-stable; `--warm-*` flip lightness in dark mode and would wash it out) plus the app's existing `feTurbulence` grain, with the June mark pinned to the top-left corner.
- **Right:** a borderless action panel with the invite link in a filled field + **inset Copy button** (full label), and a single **"Friends referred"** stat (months and friends are 1:1, so the separate months figure was redundant).

### Robustness / cleanup
- New `errorCode()` helper distinguishes the deployment-level `referrals_unavailable` 404 from transient errors, so that case shows a calm note instead of a retry that can never succeed.
- Drops the redundant copy affordances (three triggers → one), the duplicate offer row, and the dead sign-in branch from the original implementation.
- Copy reset now uses the app's standard `useEffect` + `clearTimeout` pattern.

## Verification
- `tsc --noEmit` clean
- `prettier --check` clean
- UI to be eyeballed in-browser (light + dark)

### DIFF
--- a/src/components/sidebar/Sidebar.tsx
+++ b/src/components/sidebar/Sidebar.tsx
@@ -53,7 +53,7 @@ import {
   listHermesSessions,
   sessionTimestamp,
 } from "../../lib/hermes-adapter";
-import { messageFromError } from "../../lib/errors";
+import { errorCode, messageFromError } from "../../lib/errors";
 import { NOTE_DND_MIME } from "../../lib/dnd";
 import { useForcedEmptyStates } from "../../lib/empty-states-demo";
 import { isPrimaryShortcut, primaryShortcutLabel } from "../../lib/platform";
@@ -64,6 +64,7 @@ import type {
   ReferralSummary,
 } from "../../lib/tauri";
 import { osAccountsReferralSummary } from "../../lib/tauri";
+import { JuneMark } from "../account/AccountGate";
 import { type SettingsTab } from "../settings/AppSettings";
 import { ConfirmDialog } from "../ui/ConfirmDialog";
 import { Dialog } from "../ui/Dialog";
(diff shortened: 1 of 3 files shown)


andrewwashuta:

**Addressed**
- **P1 — focus trap broken in loading/unavailable/error state** → Removed `initialFocusSelector=".referral-copy-inset"` from the dialog. Since the copy button isn't in the DOM during the (normal) loading state, the selector missed and the close-button fallback only runs when the prop is absent — so nothing got focused. The Dialog now focuses the close button on open by default; the copy button is still reachable by Tab once the summary loads.
- **P2 — `outline: none` on `:focus-visible` for the invite-link input** → Kept `:focus` suppressed (mouse) but restored a visible keyboard ring on `:focus-visible` (`2px solid var(--ring-focus)`, `outline-offset: -2px`, clipped to `--r-sm`).

Also note: this branch was behind `main`, so I merged `main` in and resolved the one conflict in `app.css` — #319's referral-overflow patch touched the old classes this PR's dialog rewrite removes; the overflow intent is preserved via `min-width:0` on the link input and `minmax(0,1fr)` split columns. `tsc`, prettier, and the full vitest suite (523 tests) pass.

Please take another look.

PR #329, opened 2026-06-17
Polish tab chrome clipping

andrewwashuta:
## Summary
- Round the docked agent composer dissolve so it no longer paints over the main panel's bottom-left corner.
- Keep the first titlebar tab aligned with the white panel while preserving paint room at the clipped edge.
- Tune titlebar tab sizing so centered tabs keep the same visual rhythm against the card, and add focused CSS regression coverage.

## Verification
- pnpm test src/test/tab-bar.test.tsx src/test/dialog-layering.test.ts
- pnpm run lint
- pnpm exec prettier --check src/styles/app.css src/test/tab-bar.test.tsx tasks/lessons.md

### DIFF
--- a/src/components/sidebar/Sidebar.tsx
+++ b/src/components/sidebar/Sidebar.tsx
@@ -1730,7 +1730,7 @@ function ReferralDialog({
                   ) : (
                     <IconClipboard size={14} />
                   )}
-                  {copied ? "Copied" : "Copy invite link"}
+                  {copied ? "Copied" : "Copy"}
                 </button>
               </div>
               {copyError ? (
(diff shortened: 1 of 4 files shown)


andrewwashuta:

**Addressed**
- (Greptile) `cssRuleFor` used a brittle non-greedy regex and only escaped dots -> replaced the rule-body capture with a selector match that escapes regex-special characters and a small brace scanner that respects quoted braces.

Please take another look.

andrewwashuta:
@greptileai I've pushed one more small follow-up.

**Addressed**
- (Greptile) `cssRuleFor` used a brittle non-greedy regex and only escaped dots -> replaced the body capture with a selector match that escapes regex-special characters and a brace scanner that respects quoted braces.
- (Follow-up copy tweak) Referral dialog copy action now reads `Copy` / `Copied` instead of `Copy invite link` / `Copied`.

Please take another look.

PR #345, opened 2026-06-18
Improve transcription retry resilience

andrewwashuta:
## Summary
- reduce final turn transcription concurrency and retry transient Scribe failures with exponential backoff, jitter, and Retry-After support
- keep exhausted invalid-response turn failures visible as timestamped transcript gaps instead of silently hiding missing audio
- return typed Scribe API timeout envelopes and include Retry-After on transient authorization denials

## Verification
- cargo test --manifest-path src-tauri/Cargo.toml
- cargo test --manifest-path scribe-api/Cargo.toml --all-targets --all-features --locked
- pnpm run lint
- pnpm test

### DIFF
--- a/scribe-api/crates/api/Cargo.toml
+++ b/scribe-api/crates/api/Cargo.toml
@@ -18,7 +18,7 @@ scribe-services.workspace = true
 serde.workspace = true
 serde_json.workspace = true
 thiserror.workspace = true
-tower.workspace = true
+tower = { workspace = true, features = ["timeout", "util"] }
 tower-http.workspace = true
 tracing.workspace = true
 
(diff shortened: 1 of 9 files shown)


andrewwashuta:
@greptileai I've pushed changes addressing the Greptile review.

**Addressed**
- (Greptile) Retry-After retries woke in lockstep. `transient_retry_delay` now treats `Retry-After` as the server floor and adds deterministic jitter on top.
- (Greptile) Invalid-response gap detection was coupled to friendly English copy. `NoteEditor` now uses the raw-message `isInvalidScribeResponseMessage` predicate exported from `NoteFailureBanner`, with the regression test updated to cover the raw `scribe_api_response_invalid` code.

Please take another look.

andrewwashuta:
@greptileai I've pushed one Codex-pass change after your 5/5 review.

**Addressed**
- (Codex) Client-side reqwest timeouts were classified the same as server timeout envelopes. `is_retryable_transcription_error` now retries the server's exact `timeout` envelope but does not retry a client-side `operation timed out` message, with a regression test covering both cases.

Please take another look to confirm this did not regress the prior Greptile sign-off.

PR #351, opened 2026-06-19
Prevent recorder status poll backlog

andrewwashuta:
## Summary
- Make recorder status polling single-flight so slow status IPC calls cannot stack up during long recordings.
- Add a regression test that keeps one status request pending and verifies later ticks do not start overlapping polls.

## Tests
- pnpm vitest run src/test/app-notes-reliability.test.tsx
- pnpm run lint
- pnpm test


andrewwashuta:

**Addressed**
- Greptile: Test did not verify polling resumes after the in-flight call settles -> updated the regression test to keep the second status request pending and assert a second poll fires after the first request resolves.

Please take another look.

PR #394, opened 2026-06-22
Fix Scribe charge replay handling

junhohong:
## Summary
- include the preview audio SHA-256 digest in Scribe preview charge idempotency keys so repeated legacy preview note IDs do not collapse different chunks into one OS Accounts charge key
- retry `/charge` response-body read failures with the same action token and idempotency key, allowing OS Accounts to return the cached settlement instead of surfacing a transient network failure
- share the existing SHA-256 hex helper between agent chat and transcription metering

## Test plan
- `cd scribe-api && cargo test -p scribe-services note_transcribe_preview -- --test-threads=1`
- `cd scribe-api && cargo test -p scribe-providers charge_ -- --test-threads=1`
- `cd scribe-api && cargo test --all-targets --all-features --locked -- --test-threads=1`
- `cd scribe-api && cargo clippy --all-targets --all-features --locked -- -D warnings`

### DIFF
--- a/scribe-api/crates/providers/src/os_accounts.rs
+++ b/scribe-api/crates/providers/src/os_accounts.rs
@@ -7,10 +7,10 @@ use serde::{Deserialize, Serialize};
 use std::time::Duration;
 
 const ERR_INSUFFICIENT_CREDITS: i64 = 4301;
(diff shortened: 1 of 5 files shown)


junhohong:
@codex review

@greptile-apps please review this PR when available.

PR #397, opened 2026-06-22
Fix tab strip flash during held sidebar drag

andrewwashuta:
## What

The titlebar tabs flashed / jumped left during a **held** sidebar drag (drag the sidebar off the left edge, hold, drag back). This drives every sidebar-width-dependent offset from a single animating value so they can't drift apart mid-snap.

## Why it flashed

The tab's x position was the **sum of two independently-clocked CSS transitions**:
- the grid column (`grid-template-columns`), retargeted on every pointer move during the drag-back, and
- the `.tab-bar` `padding-left`, a one-shot tween to 0 that is *not* retargeted.

When the grid stalled for a frame between retargets while the padding kept racing to 0, their sum moved **backward** — the tab snapping toward the window controls, then catching up. The card never flashed because its only off-grid term is a tiny 8px margin; the tab's was ~130px. Prior attempts kept this two-clock setup, so the drift survived.

Measured reopen trail (WebKit, the Tauri engine):
- before: `132 … 138, 178, 141, 190 …` (jumps to 178, snaps back to 141)
- after: `132 … 185, 231, 257, 280 …` (monotonic, 0px backward)

## The fix

- Register `--sidebar-w-current` as an animatable `@property` `<length>`.
- Derive the grid columns, card gutter, tab strip, composer, editor-footer and resize handle from that one value via `max(width, gutter)` calc.
- The snap transitions that single variable instead of seven separate properties; removed all the per-state override rules and per-element `left`/`padding` transitions (net −60 lines).
- Pin the variable to 0 when collapsed (the grid now reads it directly).
- Freeze the tab strip's adaptive layout (label/overflow churn) while resizing so tab content doesn't reflow mid-drag.

Resting positions are unchanged (collapsed card gutter, tab clearing the controls, expanded flush to the sidebar) — it's a lockstep fix, not a reposition.

## Verification

- Real WebKit (headless, driving the running app): reopen and collapse both monotonic, **0px backward jump** (was ~21px).
- `pnpm test` — 563 pass · `pnpm run build` · `pnpm run lint` clean.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -2586,7 +2586,10 @@ export function App() {
       data-sidebar-transition={sidebarTransition}
       style={
         {
-          "--sidebar-w-current": `${sidebarWidth}px`,
+          // The grid columns read this directly, so collapsed must pin it to 0
+          // (the stored width is preserved for the next expand). During a drag
+          // the resize logic overrides it imperatively.
+          "--sidebar-w-current": `${sidebarCollapsed ? 0 : sidebarWidth}px`,
         } as CSSProperties
       }
     >
@@ -2735,6 +2738,7 @@ export function App() {
           onClose={closeTab}
           onCloseOthers={closeOtherTabs}
           onNew={openNewChatTab}
+          layoutFrozen={sidebarResizing}
           onDragRegionPointerDown={handleTitlebarPointerDown}
         />
         <section className="main-panel">
--- a/src/app/sidebar-resize.ts
+++ b/src/app/sidebar-resize.ts
@@ -29,17 +29,6 @@ export function handleSidebarResizeStart(
   onStart();
   const handle = event.currentTarget;
   const shell = handle.closest(".app-shell") as HTMLElement | null;
-  const mainPanel = shell?.querySelector(".main-panel") as HTMLElement | null;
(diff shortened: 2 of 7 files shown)


andrewwashuta:
@greptileai Pushed changes addressing the review.

**Addressed**
- `initial-value` couples to `--sidebar-w` (tokens.css) → added a comment on both the `@property initial-value` and `--sidebar-w` noting they must stay in sync.
- Collapsed tab clearance → documented `--titlebar-tabs-clearance` as the intended collapsed resting offset.

**Re: the ~18px collapsed tab shift (intentional)**
Good catch, and you're right it shifts vs `main`. It's deliberate: this PR bundles the earlier `--sp-2` → `--sp-8` clearance bump so the collapsed tab pill clears the sidebar-toggle glyph instead of nearly overlapping it (the toggle occupies ~x[80,108]; old 6px gap put the pill at ~114, new 24px gap puts it at 132). Keeping `--sp-8`. My "resting positions unchanged" note was only true vs the pre-refactor working tree, not vs `main` — apologies for the imprecision.

The single-clock animation fix itself leaves expanded positions identical and makes the collapse/reopen lockstep (verified in WebKit: 0px backward jump on reopen).

Please take another look.

junhohong (approved):

PR #406, opened 2026-06-23
Fix issue report attachment delivery

0xgaut:
## Summary
- mark OS Platform issue-report uploads as public attachment files so they can be attached to created issues
- retry issue creation without `file_ids` when OS Platform still rejects attached files as private, preserving the report instead of logging-only fallback
- add a regression test for the `file must be public to attach` response

## Live verification
- Replayed the recovered June reports and confirmed they filed once each with attached screenshots: JUN-74, JUN-75, JUN-76, JUN-77, JUN-78
- Removed earlier no-file replay copies and temporary smoke/test issues so there are no duplicate report copies in `june/bug-reports`

## Test plan
- `cargo fmt --all`
- `cargo test -p scribe-providers issue_reports -- --nocapture`

## Notes
The fallback remains best-effort: if OS Platform rejects file attachment for any reason matching the private-file error, the issue still files without files and logs the fallback.

### DIFF
--- a/scribe-api/crates/providers/src/issue_reports.rs
+++ b/scribe-api/crates/providers/src/issue_reports.rs
@@ -94,7 +94,10 @@ impl OsPlatformIssueReportSink {
                     continue;
                 }
             };
-            let form = reqwest::multipart::Form::new().part("file", part);
+            let form = reqwest::multipart::Form::new()
+                .text("is_public", "true")
+                .text("purpose", "attachment")
+                .part("file", part);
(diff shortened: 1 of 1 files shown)


0xgaut:
@greptileai please review this PR.

Context: this fixes June issue report attachments by uploading OS Platform files as public attachment files, and keeps a fallback retry without `file_ids` if OS Platform still rejects the files.

Tests run locally:
- `cargo fmt --all`
- `cargo test -p scribe-providers issue_reports -- --nocapture`

junhohong (approved):

PR #408, opened 2026-06-23
[fix] Persist new session drafts

wirew0lf:
## Summary
- Persists unsent Agent composer drafts for new sessions and existing sessions in local storage.
- Restores a saved new-session draft on remount instead of falling back to the last existing session.
- Clears saved drafts after submit or session deletion.

## Validation
- pnpm test src/test/agent-workspace.test.tsx
- pnpm run lint
- pnpm exec prettier --check src/components/agent/AgentWorkspace.tsx src/test/agent-workspace.test.tsx
- pnpm test

Resolves JUN-80


junhohong (changes requested):
P1's to address from greptile

PR #411, opened 2026-06-23
Filter diagnostic artifact mentions

KaoRz:
## Summary
- Filter filename-only generated-file cards so diagnostic mentions do not surface scratch files.
- Keep genuine generated outputs visible when response text describes a file as saved, available, written, or downloadable.
- Add regression coverage for internal context files, screenshot variants, and mixed diagnostic plus real-output responses.

## Root cause
The session file UI built generated-file cards from the full workspace snapshot and treated any filename mention in a response as a surfaced output. That made diagnostic text about files like `june-context.md` or resized screenshot variants appear as downloadable generated files.

## Validation
- `pnpm run lint`
- `pnpm test`
- `pnpm run build`

Fixes JUN-77

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -7642,6 +7642,23 @@ function filesystemEntriesToArtifacts(
   });
 }
 
+const ASSISTANT_ARTIFACT_SURFACE_RE =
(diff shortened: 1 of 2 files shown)


junhohong (changes requested):
Duplicate of #407?

PR #413, opened 2026-06-23
Make report submissions fee-free

KaoRz:
## What changed
- Send bug reports, feedback, and feature requests directly through the existing issue-report endpoint instead of starting a metered chat session.
- Preserve report category, description, and attachments in the direct submission flow, with retry restoration on failure.
- Remove the obsolete report prompt wrapper and queued report-session delivery path.

## Validation
- `pnpm test src/test/agent-workspace.test.tsx src/test/issue-report-prompt.test.ts`
- `pnpm run lint`


junhohong (changes requested):
Wrong impl; reports still need to go through the LLM, just that it should be free.

PR #414, opened 2026-06-23
Keep Hermes runs active through tool continuations

KaoRz:
## Summary

- Keep Hermes gateway streams active after `message.complete` so later tool events can still arrive.
- Treat persisted histories ending in tool results as unfinished when there is no final assistant response after the latest user prompt.
- Keep live tool events on the current turn while preventing follow-up prompts from attaching tool activity to older responses.

## Issue

JUN-78

## Validation

- `pnpm test`
- `pnpm lint`
- `pnpm build`


junhohong (changes requested):
fix

PR #418, opened 2026-06-23
Fix per-chat agent model selection

andrewwashuta:
## Summary
- Scope composer model changes to the active Hermes chat by saving selected models on session metadata.
- Keep the generation model setting as the default for new chats only.
- Preserve locally known session models when refreshing session lists.

## Validation
- pnpm run lint
- pnpm test src/test/agent-workspace.test.tsx
- pnpm test

### DIFF
--- a/src-tauri/src/scribe_api.rs
+++ b/src-tauri/src/scribe_api.rs
@@ -382,10 +382,18 @@ fn normalize_agent_chat_request_for_proxy(body: &mut serde_json::Value) {
     let Some(object) = body.as_object_mut() else {
         return;
     };
-    object.insert(
-        "model".to_string(),
-        serde_json::Value::String(crate::providers::generation_model()),
-    );
+    let has_request_model = object
+        .get("model")
+        .and_then(serde_json::Value::as_str)
+        .map(str::trim)
+        .map(|model| !model.is_empty())
+        .unwrap_or(false);
+    if !has_request_model {
+        object.insert(
+            "model".to_string(),
+            serde_json::Value::String(crate::providers::generation_model()),
+        );
+    }
     clamp_agent_chat_output_tokens(object, "max_tokens");
     clamp_agent_chat_output_tokens(object, "max_completion_tokens");
 }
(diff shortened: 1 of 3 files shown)


andrewwashuta:
@greptileai please review this PR.

andrewwashuta:

**Addressed**
- (Greptile) Removed the unused local `has_model_config` write; session model state now relies only on `session.model`, matching the rest of the component logic.

Please take another look.

andrewwashuta:

**Addressed**
- (Greptile) Backfilled model-less sessions during the session merge path as well as the state effect, so sessions that load after the default model no longer remain model-less and later follow global default changes.
- (Greptile) Added local per-session model overrides for explicit picker changes, so stale background refreshes cannot overwrite the user's selected model while the session update is settling.
- Added regression coverage for late-loaded existing chats and stale server model refreshes.

**Validation**
- pnpm run lint
- pnpm test src/test/agent-workspace.test.tsx
- pnpm test
- git diff --check

Please take another look.

andrewwashuta:
@greptileai I've pushed one Codex-pass change after your 5/5 review.

**Addressed**
- (Codex) Preserved a non-empty Hermes/session-supplied `model` in the `/v1/chat/completions` proxy instead of always overwriting it with the global generation model. Added Rust tests for preserving a session model and falling back to the configured model when missing or blank.

**Skipped (with reason)**
- (Codex) Persist existing-session model changes through `PATCH /api/sessions/{id}`. I checked the pinned Hermes runtime source (`2bd1977d8fad185c9b4be47884f7e87f1add0ce3`): that PATCH route currently accepts only `title`, `archived`, and `profile`, not `model`, so sending model there would fail against the bundled runtime. The inference path now honors the request model, and new-session creation already forwards the model.

**Validation**
- `cargo test agent_proxy_ --lib`
- `cargo fmt --check`
- `pnpm run lint`
- `pnpm test`
- `git diff --check`

Please take one final look.

andrewwashuta:
@junhohong

junhohong (approved):

PR #439, opened 2026-06-25
Fix Accessibility permission refresh after grant

junhohong:
Closes JUN-95

## What changed
- Route the Accessibility banner through the app-level grant handler instead of opening System Settings directly from the banner.
- Stop opening System Settings programmatically immediately after calling the native Accessibility prompt, so macOS does not leave the Open System Preferences / Deny prompt behind the settings handoff.
- Start a bounded permission-status refresh loop after the user asks to grant Accessibility, so the app notices the grant even if the webview focus event is missed.
- Apply the same native-prompt handoff behavior to onboarding's Accessibility row.
- Add a regression test for the missing-accessibility banner flow.

## Why
Granting Accessibility happens outside June in macOS System Settings. The previous flow could both show the native TCC prompt and open Settings itself, leaving the native prompt stale after access was granted. The app also only had a one-shot focus refresh in the main flow.

## Validation
- pnpm test src/test/app-shortcut.test.tsx
- pnpm test src/test/onboarding.test.tsx
- pnpm run lint

## Known gaps
- Not manually reproduced on macOS 26.5.1 in this session.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -193,6 +193,8 @@ const AGENT_MENU_BAR_SESSION_LIMIT = 6;
 const AGENT_MENU_BAR_SESSION_RETRY_DELAYS_MS = [
   250, 500, 1000, 2000, 4000, 8000,
 ];
+const ACCESSIBILITY_PERMISSION_REFRESH_INTERVAL_MS = 1000;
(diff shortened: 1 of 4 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please review this PR when available.

PR #440, opened 2026-06-25
Fix bug report chip seeding

junhohong:
Closes JUN-109

## What changed
- Seed the report category chip synchronously for already-mounted agent composers, so the first bug-report action attaches the tag immediately.
- Keep mount-time pending-marker hydration deferred to avoid editor lifecycle warnings.
- Added a regression test that dispatches the bug-report category event against an open composer and asserts the chip appears before any timer tick.

## Why
The report flow can deliver the same fresh-session request through session storage and a window event. The previous category seed always waited for a zero-delay timer, which left a timing window where the first visible press could appear to do nothing until another action refreshed the composer state.

## Validation
- pnpm test -- src/test/agent-workspace.test.tsx
- pnpm run lint

## Known gaps
- Full app build not run; this is a scoped frontend state-management fix covered by the focused test file and TypeScript check.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -2664,7 +2664,9 @@ export function AgentWorkspace({
 
     const pending = pendingNewSessionRequest();
     if (pending) {
-      void windowEventHandlersRef.current.startNewTask(pending);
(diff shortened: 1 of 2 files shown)


junhohong:
@codex review

PR #446, opened 2026-06-25
JUN-86 Add composer slash commands

junhohong:
Closes JUN-86

## What changed
- Added built-in composer slash command parsing for /model and /file.
- Wired /model to the existing generation model switch path, including live-session command.dispatch when a chat is open.
- Wired /file to the existing Hermes workspace file import path, including quoted typed paths and the existing picker fallback when no path is supplied.
- Added Model and File to the existing slash palette for keyboard discovery.

## Why
JUN-86 asks for keyboard-driven model switching and file attachment so users can avoid leaving the composer for core chat actions.

## Validation
- pnpm exec vitest run src/test/agent-composer-slash-commands.test.ts src/test/composer-editor-slash-menu.test.tsx src/test/agent-workspace.test.tsx
- pnpm run lint
- pnpm test
- pnpm run build

## Known gaps
- /file accepts typed paths and quoted paths, then leaves prompt composition to the next message. It does not combine a file path and freeform prompt in the same slash command.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -241,6 +241,13 @@ import {
   skillSlashResolutionError,
 } from "../../lib/skill-slash-commands";
 import {
+  isBuiltinComposerSlashCommand,
+  parseBuiltinComposerSlashCommand,
+  parseSlashFileArguments,
+  resolveSlashModel,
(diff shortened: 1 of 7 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please run another review pass on the updated commit when available.

PR #447, opened 2026-06-25
Fix agent message turn ordering

junhohong:
Closes JUN-105

## What changed
- Anchor optimistic Hermes user messages to the last persisted message known at send time, then merge pending messages back into the display list at that anchor.
- Add deterministic chat-turn sorting for same-timestamp user/assistant messages and pending-user versus coarse assistant timestamps.
- Add regression coverage for the fetched-assistant-before-user race and same-timestamp turn ordering.
- Stabilize the existing issue-report retry test so it asserts the sending state while the submit promise is actually pending.

## Why
Hermes can return or stream an assistant reply before the matching user row is persisted, and Hermes timestamps can be coarse. The previous UI displayed `persisted + pending` and sorted mostly by timestamp, so the assistant reply could render above the pending user question.

## Validation
- `pnpm run lint`
- `pnpm test src/test/agent-chat-runtime.test.ts src/test/agent-workspace.test.tsx`
- `pnpm test`

## Known gaps
- No manual Tauri app smoke test run; this is covered by the frontend runtime and workspace regression tests.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -1352,6 +1352,9 @@ export function AgentWorkspace({
   const [hermesSessionMessages, setHermesSessionMessages] = useState<
     Record<string, HermesSessionMessage[]>
   >({});
+  const hermesSessionMessagesRef = useRef<
+    Record<string, HermesSessionMessage[]>
+  >({});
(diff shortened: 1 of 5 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please run a review pass on this PR when available.

PR #450, opened 2026-06-25
Fix Hermes chunk spacing in visible messages

junhohong:
## Task

Closes JUN-30.

## What changed

- Reused the runtime Hermes content extractor for visible session message text in the workspace.
- Removed the duplicate workspace extractor that trimmed each chunk before joining arrays, which could turn chunked text like "full" + " details" into "fulldetails" in issue-report diagnostics, title prompts, and pending-message comparisons.
- Added regression coverage for chunk boundary spaces in issue report diagnostics and session title suggestion prompts.

## Why

The transcript renderer already preserves chunk boundary whitespace through `textFromHermesContent`, but the workspace had a second local extractor for non-rendered visible-message uses. That extractor trimmed nested string chunks and could preserve the same kind of concatenated words shown in the JUN-30 screenshot when those paths read Hermes messages.

## Validation

- `pnpm test -- src/test/agent-workspace.test.tsx`
- `pnpm test -- src/test/agent-chat-runtime.test.ts`
- `pnpm run lint`

## Known gaps

- The live platform issue data could not be fetched locally because `OS_PLATFORM_API_KEY` is not set, so the build used the supplied screenshot as the source of truth.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -270,6 +270,7 @@ import {
   buildAgentChatTurns,
   buildHermesSessionChatTurns,
   repairContractionSpacing,
+  textFromHermesContent,
   type AgentApprovalChoice,
   type AgentChatPart,
   type AgentChatTurn,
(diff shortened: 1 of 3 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please run a review pass on this PR when available.

PR #451, opened 2026-06-25
Title issue reports from the diagnosis the model already writes

junhohong:
Closes JUN-57

Supersedes #448.

## What changed
- Updated the report investigation prompt (`src/lib/issue-report-prompt.ts`) so June opens **every** report diagnosis with an `Issue 1: <short title>` heading, where `<short title>` is a concise sentence-case tracker title. Multi-issue reports keep adding `Issue 2:`, `Issue 3:`, etc.
- In scribe-api, single-issue reports now lift that `Issue 1:` heading as the os-platform Issue title (`diagnosis_issue_title`), falling back to the existing description-derived title when no heading is present.
- Added Rust tests for the heading extraction and the single-issue title path, plus a frontend test asserting each preamble instructs starting with the `Issue 1:` heading.

## Why
Issue reports were titled from the first useful line of the user's raw report, which often preserves user phrasing instead of a concise tracker title.

Instead of #448's approach (a second, dedicated cheap Venice "title model" call), this reuses **the same model** that already writes the diagnosis: we just ask it, in the original prompt, to also produce a reasonable title. No extra model, config (`title_model`), `.env`, docker-compose, or upstream wiring — the title rides along with the diagnosis we already generate.

## Notes
- A diagnosis that splits into more sections than the cap (`MAX_SPLIT_ISSUES`) files as one combined issue and keeps the description-derived title, so its first heading does not undersell the rest.
- Multi-issue split behavior and the heading parser are unchanged.

## Validation
- `cargo test -p scribe-providers issue_reports --locked -- --test-threads=1` (29 passed)
- `cargo fmt -p scribe-providers -- --check` and `cargo clippy -p scribe-providers --all-targets --locked` (clean)
- `vitest run src/test/issue-report-prompt.test.ts` (6 passed) and `tsc --noEmit`, run against a node_modules symlink to the main checkout


junhohong:
Thanks @greptile-apps — addressed all three in `1ca2d32`:

- **P1 (`trim_common_diagnosis_tail` dropping step-4 notes):** Real risk. The step-4 wording change wasn't needed for the title feature, so I reverted step 4 in all three preambles back to the original global trailing note and kept only the step-3 `Issue 1: <short title>` change. This restores the existing, tested multi-issue split behavior; the per-section truncation risk is gone.
- **P2 (`split_issues.is_empty()` conflates 0/1-heading):** Clarified the comment — it intentionally covers both the single-heading and heading-less plain-prose cases (the latter falls back via `diagnosis_issue_title` returning `None`).
- **P2 (test only had `Issue 1:` as the first line):** Updated both fixtures to put step-1/step-2 prose before the heading, matching real diagnosis output and exercising the `find_map` scan past the preamble.

Local: `cargo test -p scribe-providers issue_reports` (29 passed), fmt/clippy clean, vitest (6 passed), `tsc --noEmit` clean. Requesting a re-review.

PR #454, opened 2026-06-26
JUN-114: Reduce message action control height

jakubswierczek:
Closes JUN-114

## What changed
- Reduced the message action row reserved top margin from `--sp-1` to `--sp-px`.
- Reduced per-action padding to compact spacing tokens.
- Added a raw CSS regression test for the compact hidden controls.

## Why
Hidden controls use opacity for hover reveal, so they still occupy flow space when not visible. Tightening the row and action padding reduces non-hovered message spacing without changing behavior.

## Validation
- `./node_modules/.bin/vitest run src/test/agent-turn-actions-css.test.ts`
- `./node_modules/.bin/tsc --noEmit`
- `./node_modules/.bin/prettier --check src/styles/app.css src/test/agent-turn-actions-css.test.ts`

## Known gaps
- Used direct local binaries because `pnpm exec` attempted to hydrate dependencies with pnpm 11 and stopped at its build approval gate before running Vitest.

### DIFF
--- a/src/styles/app.css
+++ b/src/styles/app.css
@@ -1670,7 +1670,7 @@ input:focus-visible,
   display: flex;
   gap: var(--sp-1);
   width: 100%;
-  margin-top: var(--sp-1);
+  margin-top: var(--sp-px);
   opacity: 0;
   transition: opacity 0.12s ease-out;
 }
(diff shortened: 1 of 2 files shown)


jakubswierczek:
@codex review

PR #456, opened 2026-06-26
JUN-113: Fix branching from delegated session parts

jakubswierczek:
Closes JUN-113

## What changed
- Threaded an optional source session id through the Branch from here action.
- Derived the branch source from session-bearing approval, clarify, sudo, and secret parts so delegated subagent turns call session.branch with the delegated session id.
- Kept the selected parent session as the fallback and mode-routing source.

## Why
Delegated prompt parts can belong to a subagent session even while the parent session is selected. Branching with the parent id can make Hermes reject the request with session not found.

## Validation
- ./node_modules/.bin/vitest run src/test/hermes-session-branch.test.tsx
- ./node_modules/.bin/tsc --noEmit
- git diff --check


jakubswierczek:
@codex review

PR #457, opened 2026-06-26
Fix Hermes new-session bridge race

jakubswierczek:
Closes JUN-116

## What changed

- Wait for `ensureHermesBridgeSession` before migrating a new optimistic Hermes session to its persisted session id.
- Preserve the existing bridge-session ensure path for sends into already persisted sessions.
- Add a regression test that holds bridge-session persistence open and verifies the message fetch does not target the persisted id early or surface the transient 404.

## Why

The selected-session messages effect starts fetching as soon as `selectedHermesSessionId` switches to the persisted id. For new-session sends, that switch happened before the bridge HTTP session had been persisted, so the fetch could briefly hit a missing session and show a 404 banner.

## Validation

- `./node_modules/.bin/vitest run src/test/agent-workspace.test.tsx` (131 passed, 2 skipped)
- `./node_modules/.bin/tsc --noEmit`
- `git diff --check`

## Notes

- `pnpm test -- src/test/agent-workspace.test.tsx` and `pnpm run lint` were attempted, but pnpm stopped during its install preflight on the ignored-builds approval gate for `central-icons`, `esbuild`, and `unicode-animations`. The direct local binary commands above validated the focused test suite and the lint-equivalent TypeScript check without changing dependency policy.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -3909,7 +3909,17 @@ export function AgentWorkspace({
       rememberSessionMode(storedSessionId, fullModeDraftRef.current);
     }
     const sessionDisplayTitle = sessionTitle || fallbackSessionTitle;
+    const ensureStoredHermesSession = () =>
+      withTimeout(
+        ensureHermesBridgeSession({
+          sessionId: storedSessionId,
+          title: sessionDisplayTitle,
+          ...(targetSessionModelId ? { model: targetSessionModelId } : {}),
+        }),
+        2500,
+      ).catch(() => undefined);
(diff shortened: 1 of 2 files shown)


jakubswierczek:
@codex review

jakubswierczek:
@greptile-apps Addressed the timeout-path finding in e26f026. New optimistic sessions now await the actual ensureHermesBridgeSession result before migration, and roll back on ensure failure; the existing persisted-session path keeps the bounded timeout. The regression now advances past 2.5s while persistence is still pending and verifies session-2 is not fetched early.

jakubswierczek:
@codex review

PR #461, opened 2026-06-26
Show agent action rows outside Thought

jakubswierczek:
Closes JUN-118
Closes JUN-119

## What changed
- Render tool rows in a visible stack outside the Thinking/Thought disclosure.
- Keep AgentThinkingGroup scoped to reasoning text only, so Thought contains reasoning without absorbing action rows.
- Add a regression test asserting tool rows stay outside the disclosure while reasoning stays inside.

## Why
Users need to see observable agent actions without expanding internal reasoning, while thought content stays contained under Thought.

## Validation
- ./node_modules/.bin/vitest run src/test/agent-workspace.test.tsx (143 passed, 2 skipped)
- ./node_modules/.bin/tsc --noEmit
- git diff --check

## Known gaps
- Did not run the full pnpm test suite. pnpm v11 auto-install stopped on build-script approval in the fresh worktree, so I used local binaries after the partial install.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -9101,11 +9101,11 @@ function AgentChatTurnRow({
     (part): part is Extract<AgentChatPart, { type: "tool" }> =>
       part.type === "tool",
   );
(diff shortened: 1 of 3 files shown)


jakubswierczek:
@codex review

PR #465, opened 2026-06-26
Fix push-to-talk after hands-free dictation

junhohong:
Task: user-reported bug, no tracker issue. After using Hands-free mode, push-to-talk could not be triggered anymore.

What changed:
- Reset the shortcut activation controller when direct helper listening commands bypass the shortcut down/up path.
- Added regression coverage for a hands-free toggle state being cleared before push-to-talk starts again.
- Added command classification coverage for direct stop, discard, and toggle helper commands.

Why:
- The shortcut controller intentionally blocks push-to-talk while toggle dictation is active. Stopping hands-free through the HUD sends `stop_and_paste` directly, so the controller never saw the second toggle press that would clear `active_mode`.

Validation:
- `cargo fmt --manifest-path src-tauri/Cargo.toml`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked external_hands_free_stop_allows_push_to_talk_again`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked direct_helper_listening_commands_reset_shortcut_activation`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked dictation::tests`

Known gaps:
- Not manually verified against the packaged macOS helper in a running app from this worktree.

### DIFF
--- a/src-tauri/src/dictation.rs
+++ b/src-tauri/src/dictation.rs
@@ -774,7 +774,9 @@ pub fn dictation_helper_command(
     command: serde_json::Value,
 ) -> Result<(), AppError> {
     #[cfg(target_os = "macos")]
-    let _ = &app;
+    if helper_command_resets_shortcut_activation(&command) {
+        reset_shortcut_activation(&app);
+    }
 
     #[cfg(not(target_os = "macos"))]
     if state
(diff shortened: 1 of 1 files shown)


junhohong:
@codex review

PR #468, opened 2026-06-26
Teach June to clarify ambiguous first prompts

junhohong:
Closes JUN-121

## What changed
- Added a dedicated clarifying-questions section to June's generated Hermes SOUL.md.
- Calibrated the behavior to ask one or two questions before acting when the first user message is ambiguous, expensive to guess, or has multiple reasonable interpretations.
- Kept the existing fast path for clear or low-risk requests so June does not add unnecessary friction.
- Added a Hermes bridge unit test that pins the new soul guidance.

## Why
Users reported that June can start acting immediately on underspecified initial prompts. The existing clarify surface already exists, so this teaches the runtime when to use it.

## Validation
- cargo fmt --manifest-path src-tauri/Cargo.toml
- cargo test --manifest-path src-tauri/Cargo.toml --locked hermes_bridge
- cargo test --manifest-path src-tauri/Cargo.toml --locked

## Known gaps
- No manual Hermes runtime session was exercised in this local run.


junhohong:
@codex review

PR #471, opened 2026-06-26
Disable turn action hover animation

junhohong:
## Summary
- Keep per-message action controls out of normal layout so hover reveal cannot expand the message row
- Reveal copy/edit/branch controls instantly with opacity and pointer-event toggles instead of grid-row animation
- Update the focused CSS regression test to lock in no height-changing action-row reveal

## Validation
- `pnpm exec vitest run src/test/agent-turn-actions-css.test.ts`
- `pnpm run lint`
- `pnpm exec prettier --check src/styles/app.css src/test/agent-turn-actions-css.test.ts`

## Known gaps
- No tracker issue was provided in the prompt.


junhohong:
@codex review

PR #476, opened 2026-06-27
Allow short onboarding practice replies

junhohong:
## What changed
- Treat any non-empty trimmed onboarding practice reply as a successful dictation practice response.
- Added an onboarding test proving a one-character reply enables Start using June.

## Why
Users can reasonably answer the practice prompt with very short text such as "hi". The previous four-character threshold kept the primary action disabled even though text had arrived.

## Validation
- pnpm test -- src/test/onboarding.test.tsx
- pnpm run lint

## Known gaps
- None.


junhohong:
@codex review

PR #502, opened 2026-06-29
Remove service branding from verify page

junhohong:
Closes JUN-125.

## What changed

- Removed Cargo package/service-name interpolation from the `/verify` page title, lede, and deployment facts.
- Replaced the service fact row with a neutral version row.
- Added renderer and HTTP boundary assertions so the public verify HTML no longer shows legacy `scribe-api`/`scribe api` naming or a service row.

## Why

The legacy `scribe-api.opensoftware.co/verify` deployment currently exposes stale `scribe-api` naming in the verify page copy. The verify page should describe the server and attestation flow without deriving visible copy from package or service identifiers.

## Validation

- `cargo test -p june-api verify --locked -- --test-threads=1`
- `pnpm test:june-api`

## Live walkthrough

Skipped. This is a static backend HTML endpoint, and the HTTP boundary test renders the exact public `/verify` response. A QA video would not add useful reviewer evidence for this change.

## Deployment note

The live legacy host is currently serving an older pre-rename image. This PR removes the package-name branding path in source; the host will need a current June API deployment before the live page reflects the change.

### DIFF
--- a/june-api/crates/api/src/handlers/verify.rs
+++ b/june-api/crates/api/src/handlers/verify.rs
@@ -57,7 +57,6 @@ fn render_page(info: &AttestationInfo) -> String {
     };
 
     PAGE_TEMPLATE
-        .replace("@SERVICE@", env!("CARGO_PKG_NAME"))
         .replace("@VERSION@", env!("CARGO_PKG_VERSION"))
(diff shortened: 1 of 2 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please review this PR.

PR #506, opened 2026-06-29
Prompt for Accessibility access from onboarding

junhohong:
## What changed

- Added a Tauri `request_accessibility_permission` command that calls macOS `AXIsProcessTrustedWithOptions` with the prompt option instead of only opening System Settings.
- Added a shared frontend `requestDictationAccessibilityPermission` helper used by onboarding, the Accessibility banner, and the Settings System permissions Manage button.
- Kept the dictation-helper prompt as the required path so the helper process that owns dictation paste can register with TCC.
- Added an app-process prompt only as a best-effort fallback when the helper request cannot be delivered, while still surfacing helper failure so the UI opens the Accessibility settings fallback.
- Added regressions for onboarding's Accessibility Allow button, the Settings Manage Accessibility button, and helper failure fallback behavior.

## Why

Clicking Accessibility in onboarding should ask macOS for the native Accessibility permission in context. The System permissions Manage action should take the same working path instead of silently doing nothing.

## Review feedback addressed

- Greptile flagged concurrent app/helper prompting and swallowed helper failures. Fixed by requesting the helper first, avoiding the app prompt on the normal path, and opening Accessibility settings only when the helper request fails.
- Codex reported no major issues on the first pass.

## Validation

- `pnpm test -- src/test/onboarding.test.tsx src/test/app-shortcut.test.tsx`
- `pnpm run lint`
- `pnpm test`
- `pnpm build`
- `pnpm run format`
- `git diff --check`

## Live walkthrough

BLOCKED for the actual macOS Accessibility prompt: clicking through the real prompt would mutate this Mac's TCC privacy state and requires explicit confirmation at action time. The native command was compiled in the full Tauri Rust test suite, and the user-visible paths are covered by App/onboarding tests that assert no System Settings handoff happens for the prompt path.

## QA video

Skipped. The proof point is a macOS privacy prompt, and recording/clicking it would require the same explicit OS-permission side-effect confirmation.

### DIFF
--- a/src-tauri/src/commands.rs
+++ b/src-tauri/src/commands.rs
@@ -40,6 +40,7 @@ use crate::{
     },
 };
 use chrono::{TimeZone, Utc};
+use serde::Serialize;
 use sqlx::query::query;
 use sqlx::row::Row;
 use sqlx_sqlite::SqlitePool;
@@ -53,6 +54,12 @@ use std::{
 };
 use tauri::AppHandle;
 use tokio::sync::OnceCell;
+
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename_all = "camelCase")]
+pub struct AccessibilityPermissionResponse {
+    state: String,
+}
 
 #[tauri::command]
 pub async fn bootstrap_app(app: AppHandle) -> Result<BootstrapResponse, AppError> {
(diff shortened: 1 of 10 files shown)


junhohong:
@greptile-apps please review this PR.

junhohong:
@codex review

junhohong:
@greptile-apps please review this PR again after 953b7c2. The helper prompt is now required, app prompting is fallback-only, and helper failures open the Accessibility settings fallback.

junhohong:
@codex review

PR #519, opened 2026-06-30
Redesign billing section as improved plan + usage meter

andrewwashuta:
Reworks the Account → Billing card to the Cursor-style "plan + usage meter" layout the team aligned on (percentages for usage, $ stays in Manage billing).

## What
- **Plan header over a labeled usage meter.** Replaces the stacked subscription row + floating "X% remaining" number with a quiet plan header (name, status line, primary action) above a full-width meter (`Usage remaining` label + tabular percentage + refresh).
- **Low-usage color.** Since the meter shows *remaining*, the fill warms to `--destructive` at ≤15% so green never implies a near-empty allowance.
- **Rounded meters — bug fix.** `border-radius: var(--r-full)` referenced a token that's defined nowhere, so the billing meter *and* the sibling mic-test meter were silently square. Both now use the existing `--r-pill` token.
- **`__billingDemo(...)` dev console driver** (gated to `import.meta.env.DEV`, same pattern as `__emptyStates`): parks the card in any plan state — free, pro, trialing, past due, running low, signed out — or `"all"` to stack every variant for side-by-side design review.

## Why
- Matches the agreed direction by team and keeps usage in % while subscription/top-up stay in $.
- Colors are all pre-existing system tokens interpolated in oklch; no hand-coded values.
- The card logic is unchanged for real users — derivation still keys off `hasLiveSubscription` / `billingRecovery` / balance; only the layout and the square-corner bug changed.

## Notes
- Paid tier renders as "Pro plan" via a single `PAID_PLAN_NAME` constant — one line to rename when the plan catalog grows.
- `app-settings` tests updated to the new copy/structure; all green, typecheck + prettier clean.

### DIFF
--- a/src/components/account/AccountSettings.tsx
+++ b/src/components/account/AccountSettings.tsx
@@ -213,11 +213,12 @@ export function BillingSettingsSection({
   }
 
   // __billingDemo(...) parks the card in a fixture state (or "all" stacks every
-  // variant) for design work; demoPlan stays null in production. See
-  // lib/billing-demo.ts.
-  const galleryMode = demoPlan === "all";
+  // variant) for design work. Every fixture reference is gated on
+  // import.meta.env.DEV so the bundler drops BILLING_DEMO_FIXTURES (and the six
+  // fixture accounts) from production — demoPlan is always null there anyway,
+  // since the console command is never registered. See lib/billing-demo.ts.
   const demoAccount =
-    demoPlan && demoPlan !== "all"
+    import.meta.env.DEV && demoPlan && demoPlan !== "all"
       ? BILLING_DEMO_FIXTURES[demoPlan].account
       : undefined;
   const cardProps = {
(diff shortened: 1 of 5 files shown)


andrewwashuta:
@greptileai pushed changes addressing both review observations.

**Addressed**
- `billing-demo.ts` hardcoded dates → `RENEWS_AT` / `TRIAL_ENDS_AT` are now computed relative to `Date.now()` (+24 / +7 days), so the demo always shows an upcoming renewal/trial date.
- Fixtures bundled into production → every `BILLING_DEMO_FIXTURES` / `BILLING_DEMO_ORDER` reference in `AccountSettings.tsx` is now gated behind `import.meta.env.DEV`, so the bundler drops the six fixture accounts from production builds (matches the existing `import.meta.env.DEV` pattern in `main.tsx`; `demoPlan` is always null in prod anyway since the console command is never registered).

Typecheck, the app-settings tests, and prettier are all green. Please take another look.

andrewwashuta:
@greptileai pushed another change addressing a Codex P2 (real regression, so I took it now rather than deferring).

**Addressed (Codex)**
- Subscribed accounts with `subscribed: true` but no `status` were being relabeled as `Free plan` / `No credit card required.` / Upgrade. `onPaidPlan` now keys off `subscribed === true` (with `liveSubscription` as a fallback), so partial/older subscription payloads stay on the paid plan with a `Manage billing` action; `planDetail` falls back to `Renews ...` / `Active` for the no-status case. Added a regression test (`treats subscribed accounts as paid when status is absent`).

Typecheck, app-settings tests (32), and prettier all green. Please take another look.

PR #557, opened 2026-07-01
Fix Meeting notes delete fallback

junhohong:
Closes JUN-175

## Summary
- Keep note-related views on Meeting notes when deleting leaves no notes, instead of routing to Settings.
- Apply the same empty-notes fallback to single delete, bulk delete, and failed meeting-start cleanup.
- Add App-level regressions for deleting the last note and bulk deleting every note.

## Why
JUN-175 reports that deleting the latest Meeting notes entry sends the user to Settings. The empty-list branch in the delete handlers explicitly called `setActiveView("settings")`, so the user lost the Meeting notes context after a successful delete.

## Validation
- `pnpm exec vitest run src/test/app-notes-reliability.test.tsx`
- `pnpm test`
- `pnpm build`
- `pnpm exec biome check src/app/App.tsx src/test/app-notes-reliability.test.tsx` exited 0, with existing App warnings still reported
- `git diff --check`

## Skipped or blocked
- `pnpm run check` is blocked by pre-existing repo-wide Biome diagnostics outside this change. First reported error: `scripts/hermes-smoke.ts:515` prefers a template literal; Biome also reports existing warnings in unrelated files.
- Live native QA was not run because reproducing the bug would delete real local Meeting notes. The added App regression tests drive the visible delete dialog path with mocked Tauri data and assert the empty Meeting notes page remains current.

## Known gaps
- This PR does not change how the next note is selected when notes remain. It only fixes the empty-list redirect to Settings.

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -2088,6 +2088,16 @@ export function App() {
     }
   }
 
+  function handleEmptyNotesAfterDelete() {
+    const currentView = activeViewRef.current;
+    if (currentView === "meetings" || currentView === "notes" || currentView === "all-notes") {
+      setActiveView("notes");
+    }
+    setOriginFolderId(undefined);
+    setOriginAllNotes(false);
+    setFolderReturnTarget(undefined);
+  }
+
   async function handleDeleteNote(noteId: string) {
     if (state.recordingStatus) {
       setError("Stop the current recording before deleting a note.");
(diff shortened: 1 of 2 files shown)


junhohong:
@codex review

junhohong:
@greptile-apps please run another pass on the current commit.

PR #561, opened 2026-07-01
Speed up macOS desktop release builds

junhohong:
## Summary

- Add a `macos-runner` input to the RC and promote macOS release workflows, defaulting to the Mac Studio self-hosted runner labels with a GitHub-hosted fallback.
- Restore a pinned `.tauri-hermes/hermes` cache before bundling in both signed macOS release workflows.
- Teach `scripts/bundle-hermes-runtime.sh` to reuse a matching cached Hermes bundle, re-sign Mach-O files with the current Developer ID identity, and run the relocation self-test before the app build.
- Document the Mac Studio runner labels, host setup, service setup, fallback behavior, and cache behavior.

## Validation

- `bash -n scripts/bundle-hermes-runtime.sh`
- `actionlint .github/workflows/rc-desktop-dmg.yml .github/workflows/promote-desktop.yml`
- `pnpm check` (exits 0; repo has existing Biome warnings outside this change)
- `pnpm typecheck`
- Synthetic cache-hit smoke test for `scripts/bundle-hermes-runtime.sh`
- `pnpm test` (126 files, 1856 passed, 2 skipped)
- `git diff --check`

## Notes

This keeps the release-safe behavior intact: restored Hermes runtimes are accepted only when the `PIN` matches the Rust source pin and the required runtime files exist, and they are still signed with the currently imported certificate before notarization.

### DIFF
--- a/.github/workflows/promote-desktop.yml
+++ b/.github/workflows/promote-desktop.yml
@@ -225,8 +225,6 @@ jobs:
         with:
           path: .tauri-hermes/hermes
           key: hermes-bundle-${{ runner.os }}-${{ runner.arch }}-${{ hashFiles('scripts/bundle-hermes-runtime.sh', 'src-tauri/src/hermes_bridge.rs') }}
-          restore-keys: |
-            hermes-bundle-${{ runner.os }}-${{ runner.arch }}-
 
       - name: Bundle Hermes runtime into app resources
         env:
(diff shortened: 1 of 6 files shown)


junhohong:
Addressed the automated review findings in follow-up commits:\n\n- d7e614ef restores EXIT cleanup for the Hermes relocation self-test temp directory.\n- 8a71e919 removes the broad Hermes cache restore key so release builds only reuse an exact script-plus-pin cache key.\n\nLatest CI is green, and the Mac Studio self-hosted runner is online with the expected release labels.

junhohong:
@greptileai review

PR #562, opened 2026-07-01
feat: play back and trim a recording before transcription

analoguedarkness:
## Summary

- Stopping a recording now opens a trim modal showing the clip's waveform with
  draggable start/end handles, so silence or off-topic audio can be cut from
  either end before transcription and note generation run. Dismissing the modal
  (Esc / close / "Use full recording") finalizes the full clip, so a stopped
  recording is never left unprocessed.
- The modal plays the recording back with a play/pause control and a scrubbable
  playhead over the waveform. Microphone and system sources play together (the
  browser mixes their output) so you hear what was captured; playback auditions
  exactly the kept region (starts at the start handle, stops at the end handle).
- Backend splits stop into stage/finalize phases so transcription is deferred
  until the trim decision: `prepare_recording_trim` finalizes the WAVs, stashes
  the recording, and returns a downsampled waveform plus the source paths;
  `finish_recording` applies the trim (rewriting each 16-bit PCM WAV to the
  selected window) and then runs the existing transcription pipeline.
- Security-sensitive: adds the app recordings directory to the Tauri
  asset-protocol scope at runtime (in the setup hook) so the modal's `<audio>`
  can read local WAVs. This covers both the dev (`-dev`) and prod data dirs
  without a static glob, and mirrors how `AppPaths::contained_recording_file`
  already treats everything under `recordings/` as ours.

## Validation

- `pnpm typecheck` (tsc --noEmit): clean.
- `pnpm check` (Biome): exit 0 (no errors; only the repo's pre-existing
  warnings).
- `pnpm test` (vitest): trim dialog suite (incl. playback) and the app
  trim-path suites pass.
- `pnpm test:rust` (`cargo test`): passes, including new `trim.rs` unit tests
  (`trims_to_the_selected_window`, `waveform_preview_buckets_match_timeline`).
- `cargo fmt --check` and `cargo clippy --all-targets`: clean for the changed
  files.
- Rebased onto `main` at v0.0.25 (conflicts in `App.tsx` and one app test
  resolved; branch reformatted to Biome).

- [x] Tested visually where it matters (UI changes: attach a screenshot or recording). 

## Out of scope

- No audio re-encoding or format conversion: trimming rewrites the existing
  16-bit PCM WAVs in place; non-PCM sources are left untouched.
- No waveform zoom or multi-region selection; a single start/end window only.

## Followups

- A staged-but-never-confirmed recording (e.g. app crash with the modal open)
  leaves an orphaned entry in the in-memory pending store; the WAVs remain on
  disk. Acceptable for v1 since a normal dismiss always finalizes.

### DIFF
--- a/src-tauri/src/audio/capture.rs
+++ b/src-tauri/src/audio/capture.rs
@@ -463,15 +463,29 @@ pub fn finish_capture(session_id: &str) -> Result<FinishedRecording, AppError> {
 /// later consumes it via [`take_pending_recording`].
 pub fn stage_finish_capture(session_id: &str) -> Result<FinishedRecording, AppError> {
     let finished = finish_capture(session_id)?;
-    if let Ok(mut pending) = PENDING_RECORDINGS.lock() {
-        pending.insert(finished.session_id.clone(), finished.clone());
-    }
+    // Recover from a poisoned lock rather than dropping `finished`. finish_capture
+    // has already consumed ACTIVE_RECORDING, so this staged clone is the only
+    // remaining handle to the finalized recording. Skipping the insert (or
+    // propagating an error, which drops `finished`) would strand the captured
+    // audio: `finish_recording`'s fallback would re-run finish_capture against an
+    // already-consumed session and fail.
+    pending_recordings().insert(finished.session_id.clone(), finished.clone());
     Ok(finished)
 }
 
 /// Remove a staged recording awaiting trim review, if one exists for the session.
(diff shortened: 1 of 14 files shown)


junhohong:
thank you so much for your first PR @analoguedarkness !! 

we don't yet have established standards for contributions, but in the meantime could you: 
1. Resolve Greptile/Codex reviews until clean 
2. Attach a demo recording of the feature in usage?

PR #563, opened 2026-07-01
Fix missing system audio source transcription

junhohong:
## Summary
- Add per-source full-audio fallback turn coverage after dual-source turn detection.
- Preserve existing detected turns while ensuring a kept source with zero detected turns still gets a transcription job.
- Add regression tests for the case where microphone turns exist but system audio produces no detected turns.

## Why
Prod investigation showed system audio capture and validation were succeeding, but turn detection produced only microphone turns. Because full-source fallback only ran for sources that already had turn jobs, valid system audio could be omitted from meeting notes without a system transcript or error.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --locked domain::processing::tests`
- Launched local dev app from `/Users/junhohong/code/open-software/os-june-system-audio-source-fallback` with Vite on `127.0.0.1:1421` and local June API on `127.0.0.1:8080` for manual testing.

## Live walkthrough
No agent-recorded live walkthrough yet. The affected flow records microphone and system audio, so I left the live recording test to the user-run local app rather than recording private audio without explicit confirmation.

### DIFF
--- a/src-tauri/src/domain/processing.rs
+++ b/src-tauri/src/domain/processing.rs
@@ -1819,6 +1819,46 @@ mod tests {
         assert!(covered.iter().all(|turn| turn.end_ms > turn.start_ms));
     }
 
+    #[test]
+    fn empty_detection_gets_full_source_fallback_for_every_source() {
+        let mic_path = PathBuf::from("microphone.wav");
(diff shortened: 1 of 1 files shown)


junhohong:
@greptileai review

PR #505, opened 2026-06-29
Retry duration-invalid saved recordings

junhohong:
## Summary
- let failed notes expose invalid saved audio artifacts as preserved local audio so the Retry button stays available
- make retry_processing revalidate invalid saved artifacts under the current duration rules before promoting them to valid and processing them
- keep discarded recordings hidden and keep truly unreadable or truncated invalid artifacts rejected

## Why
PR #504 fixed future recordings that produce a longer saved WAV than the app elapsed estimate, but recordings that already failed under the old duration rule may have persisted their audio artifacts as invalid. This follow-up makes Retry recover those saved artifacts when they now pass validation.

## Validation
- cargo test --manifest-path src-tauri/Cargo.toml --test notes --test source_modes --locked
- cargo test --manifest-path src-tauri/Cargo.toml --locked
- git diff --check

## Live QA
- Not run. Proving this end to end requires a long native microphone/system-audio recording or a local database fixture with private recording paths. The repository and retry behavior is covered by deterministic Rust tests.

PR #504, opened 2026-06-29
Allow longer recording audio to process

junhohong:
## Summary
- allow readable recording audio with bounded positive duration drift to continue processing
- keep blocking audio that is shorter than expected beyond tolerance, since that indicates likely truncation
- reject stale long audio for short sessions by capping positive drift to a proportional window with a 10 minute maximum
- apply the same validation decision to interrupted recording recovery and add regression coverage for accepted drift and rejected excessive overrun

## Why
A long recording can produce a saved WAV whose sample duration is greater than the elapsed duration stored by the app, especially around long sessions, helper timeline padding, drift, or pause/resume edges. The previous validation treated both longer and shorter mismatches as fatal, causing recoverable audio to fail before transcription. This keeps those recoverable cases working without allowing obviously stale long artifacts through short sessions.

## Validation
- cargo fmt --manifest-path src-tauri/Cargo.toml
- cargo test --manifest-path src-tauri/Cargo.toml --test recording_validation --locked
- cargo test --manifest-path src-tauri/Cargo.toml --locked
- git diff --check
- CI: Changed paths, Frontend lint and tests, Tauri Rust tests, Repository hygiene, and Socket checks passed on commit e736d447d5

## Reviews
- Greptile: second pass reviewed e736d447d5 with confidence 5/5 and no merge-blocking issues
- Codex: reviewed e736d447d5 and found no major issues

## Live QA
- Not run. Reproducing this exact flow requires a long native microphone/system-audio recording. The repo QA instructions require explicit confirmation before recording microphone audio. The backend regression is covered deterministically in Rust.

PR #569, opened 2026-07-01
Fix system source silently dropped from long or quiet meetings

andrewwashuta:
## Summary

- Fixes hour-long meetings where both microphone.wav and system.wav exist on disk with full duration but only the microphone is transcribed. Two upstream gates could discard the system source before the PR #563 full-source fallback ever ran.
- Gate 1: `validate_audio_artifact` now repairs stale WAV headers before reading duration, so validation judges the true recording, not a header the helper never finalized.
- Gate 2: the pre-transcription silence pre-filter now uses the same floor as system turn detection, so quiet-but-real system tracks are no longer dropped before the fallback.

## Root cause

- **Validation trusted stale WAV headers.** On a long/unhealthy capture the system-audio helper can be SIGKILLed mid-finalization (5s grace before kill). AVAudioFile only rewrites the RIFF/data chunk sizes on a clean close, so the header is left stale (smaller than the bytes actually on disk). hound's `reader.duration()` trusts that header, making an hour-long file read as minutes, which fails the `is_not_truncated` check and keeps the source out of `valid_sources` - processing proceeds mic-only. The file plays fine in tolerant players; only the header is wrong. Fix: parse the RIFF structure and, for plain 16-bit PCM, rewrite the RIFF and data size fields in place from the real byte length (either direction of mismatch) before duration is read; all existing duration/drift-tolerance rules then apply to the true duration.
- **The silence pre-filter floor was stricter than the detection floor.** `drop_silent_system_sources` ran before turn detection (and before the #563 fallback) using `SILENCE_RMS_FLOOR = 0.012`, while system-source turn detection uses `min_rms = 0.006`. A system track whose loudest 30ms window sits between 0.006 and 0.012 was detectable speech to the detector but "effectively silent" to the pre-filter, so it was dropped before the fallback could cover it. Fix: derive both floors from one shared `SYSTEM_DETECTION_MIN_RMS` constant and use it for the pre-drop; the normalized-audio chunk-skip guard keeps the 0.012 floor.

## Validation

- `cargo fmt` clean; `cargo clippy --all-targets` clean; full `cargo test` suite green (299 lib tests + all integration binaries, 0 failures).
- Added tests for both fixes; each verified to fail with the fix reverted (stale-header repair test and quiet-between-floors drop test).
- A `silent_source_dropped` session checkpoint (source + max RMS) now records any pre-filter drop for observability; a repaired header adds a non-fatal note to the validation warnings.
- Tested visually: N/A, backend-only (Rust audio processing, no UI change).
- Needs a June API (backend) deploy: no. Desktop-only change; no `/v1/*` contract touched.

## Out of scope

- The system-audio helper's stop budget (SIGTERM 5s grace before SIGKILL) is left as-is; this PR makes validation resilient to an unfinalized header rather than preventing it.
- The helper's realtime-thread I/O (root cause of glitchy capture) is untouched.
- The adaptive system detection threshold is unchanged; only the pre-filter floor is aligned to it.
- `add_full_source_turns_for_missing_sources` (#563), `coalesce_turns_for_transcription`, and the positive-drift-tolerance rules from #504 are deliberately untouched.

## Followups

- Give the helper a realistic stop budget or incremental header finalization so SIGKILL is survivable (prevention for Gate 1).
- Move capture-side work off the realtime IO thread (ring buffer + writer thread) - the actual glitchy-audio fix.
- Cap the adaptive detection threshold so dense speech cannot inflate its own noise floor (reduces reliance on the #563 fallback).
- Note for #505 (retry duration-invalid recordings): with this change, previously stale-header artifacts revalidate to their true duration, so its retry path can now recover them.

### DIFF
--- a/src-tauri/src/audio/turns.rs
+++ b/src-tauri/src/audio/turns.rs
@@ -10,10 +10,14 @@ const NORMALIZE_MAX_GAIN: f32 = 32.0;
 const TRANSCRIPTION_SAMPLE_RATE: u32 = 16_000;
 const TRANSCRIPTION_CHANNELS: u16 = 1;
 pub const MAX_TRANSCRIPTION_CHUNK_MS: i64 = 30 * 1000;
-/// Loudest-window RMS below which a track carries no transcribable speech.
-/// Deliberately conservative (≈ -38 dBFS) and matches the microphone lane's
-/// activity `min_rms`, so we only ever skip clearly-silent audio.
+/// Loudest-window RMS below which normalized audio carries no transcribable
+/// speech. Applied to already-gained audio (chunk-skip before API calls), so a
+/// track this quiet after up to 32x normalization is genuinely silent.
 const SILENCE_RMS_FLOOR: f32 = 0.012;
+/// Activity `min_rms` for the system lane's turn detection. The pre-transcription
+/// silence pre-filter must not drop below what detection can still find, so both
+/// derive from this one constant.
+pub const SYSTEM_DETECTION_MIN_RMS: f32 = 0.006;
 
 #[derive(Debug, Clone)]
 pub struct DetectionSource {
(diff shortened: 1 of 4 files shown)


andrewwashuta:
@greptileai I've pushed changes addressing the review (commit c79631a6).

**Addressed**
- Double WAV decode in the drop path (Greptile) → `partition_silent_system_sources` now decodes each system source once via `source_max_rms` and derives the silence verdict from it; unreadable files still mean "keep".
- Trailing chunks inflating the repaired data size (Greptile) → the repair now treats a declared data size as authoritative when its end lands at EOF or is followed by a plausible RIFF chunk header, and only expands when the declared end is followed by non-chunk bytes (the SIGKILL case — zero-filled or PCM tails fail the chunk-header test). Added a trailing-LIST regression test.
- 1-hour test fixture (Greptile) → shrunk to 10s claiming ~1s; same code path, ~300KB instead of ~110MB.
- Recovery expectation derived from the pre-repair header (Codex) → confirmed real: `recovery_validation_expected_duration_ms` fell through to raw `wav_duration_ms` for pending rows (expected_duration_ms = 0), so validation's own repair would then fail the source as stale-long. It now repairs the header before deriving the expectation, with a regression test.
- WAVE_FORMAT_EXTENSIBLE rejected by the repair (Codex) → confirmed in-repo precedent (`live_preview.rs` accepts `1 | 0xfffe`); `read_wav_layout` now accepts 0xFFFE with 16-bit samples, with a hand-built extensible-header test.

Nothing skipped. Full Rust suite green (302 lib tests + integration binaries, 0 failures), clippy clean.

Please take another look.

andrewwashuta:
Closed inadvertently by renaming the head branch to `jun-178-june-report-meeting-transcription-has-la` (to auto-link platform issue JUN-178). Same commits continue in #570 — Greptile had signed off 5/5 on `c79631a6` here.

PR #570, opened 2026-07-01
Fix system source silently dropped from long or quiet meetings (JUN-178)

andrewwashuta:
Closes JUN-178

Continuation of #569 (closed by a head-branch rename to link the platform issue — same commits, review history there: Greptile signed off 5/5 "safe to merge" on `c79631a6`).

## Summary

- Fixes hour-long meetings where both microphone.wav and system.wav exist on disk with full duration but only the microphone is transcribed (JUN-178: "system audio never gets brought into the conversation", 25-minute transcript gaps while the reporter was muted). Two upstream gates could discard the system source before the PR #563 full-source fallback ever ran.
- Gate 1: `validate_audio_artifact` now repairs stale WAV headers before reading duration, so validation judges the true recording, not a header the helper never finalized.
- Gate 2: the pre-transcription silence pre-filter now uses the same floor as system turn detection, so quiet-but-real system tracks are no longer dropped before the fallback.

## Root cause

- **Validation trusted stale WAV headers.** On a long/unhealthy capture the system-audio helper can be SIGKILLed mid-finalization (5s grace before kill). AVAudioFile only rewrites the RIFF/data chunk sizes on a clean close, so the header is left stale (smaller than the bytes actually on disk). hound's `reader.duration()` trusts that header, making an hour-long file read as minutes, which fails the `is_not_truncated` check and keeps the source out of `valid_sources` - processing proceeds mic-only. The file plays fine in tolerant players; only the header is wrong. Fix: parse the RIFF structure and, for plain or extensible 16-bit PCM, rewrite the RIFF and data size fields in place from the real byte length before duration is read; declared sizes corroborated by a valid trailing chunk are treated as authoritative and never expanded.
- **The silence pre-filter floor was stricter than the detection floor.** `drop_silent_system_sources` ran before turn detection (and before the #563 fallback) using `SILENCE_RMS_FLOOR = 0.012`, while system-source turn detection uses `min_rms = 0.006`. A system track whose loudest 30ms window sits between 0.006 and 0.012 was detectable speech to the detector but "effectively silent" to the pre-filter, so it was dropped before the fallback could cover it. Fix: derive both floors from one shared `SYSTEM_DETECTION_MIN_RMS` constant and use it for the pre-drop; the normalized-audio chunk-skip guard keeps the 0.012 floor.

## Review feedback already addressed (from #569)

- Drop path decodes each system WAV once (was twice) - Greptile.
- Header repair only expands to EOF when the declared end is not corroborated by a trailing RIFF chunk - Greptile.
- 1-hour (~110MB) test fixture shrunk to seconds - Greptile.
- Crash recovery repairs the stale header before deriving its expected duration, so the repair cannot make recovery worse - Codex, verified.
- Repair accepts WAVE_FORMAT_EXTENSIBLE 16-bit PCM, matching the live-preview parser - Codex, verified.

## Validation

- `cargo fmt` clean; `cargo clippy --all-targets` clean; full `cargo test` suite green (302 lib tests + all integration binaries, 0 failures).
- New tests: stale-small header repair, oversized-header truncation, trailing-chunk no-repair, extensible-format repair, recovery expectation repair, between-floors system source kept, truly-silent source dropped. Load-bearing tests verified to fail with the fix reverted.
- A `silent_source_dropped` session checkpoint (source + max RMS) now records any pre-filter drop for observability; a repaired header adds a non-fatal note to the validation warnings.
- Tested visually: N/A, backend-only (Rust audio processing, no UI change).
- Needs a June API (backend) deploy: no. Desktop-only change; no `/v1/*` contract touched.

## Out of scope

- The system-audio helper's stop budget (SIGTERM 5s grace before SIGKILL); this PR makes validation resilient to an unfinalized header rather than preventing it.
- The helper's realtime-thread I/O (root cause of glitchy capture, likely behind JUN-178's provider-rejected chunk at 9:28).
- The adaptive system detection threshold (only the pre-filter floor is aligned to it).
- The single provider-rejected mic chunk in JUN-178 ("could not process this audio") - separate failure mode, should be split into its own issue.

## Followups

- Give the helper a realistic stop budget or incremental header finalization so SIGKILL is survivable (prevention for Gate 1).
- Move capture-side work off the realtime IO thread (ring buffer + writer thread) - the actual glitchy-audio fix.
- Cap the adaptive detection threshold so dense speech cannot inflate its own noise floor.
- Note for #505: with this change, stale-header artifacts revalidate to their true duration, so its retry path can now recover them - JUN-178's original recording should be recoverable via Retry on a build with this fix.

### DIFF
--- a/src-tauri/src/audio/turns.rs
+++ b/src-tauri/src/audio/turns.rs
@@ -10,10 +10,14 @@ const NORMALIZE_MAX_GAIN: f32 = 32.0;
 const TRANSCRIPTION_SAMPLE_RATE: u32 = 16_000;
 const TRANSCRIPTION_CHANNELS: u16 = 1;
 pub const MAX_TRANSCRIPTION_CHUNK_MS: i64 = 30 * 1000;
-/// Loudest-window RMS below which a track carries no transcribable speech.
-/// Deliberately conservative (≈ -38 dBFS) and matches the microphone lane's
-/// activity `min_rms`, so we only ever skip clearly-silent audio.
+/// Loudest-window RMS below which normalized audio carries no transcribable
+/// speech. Applied to already-gained audio (chunk-skip before API calls), so a
+/// track this quiet after up to 32x normalization is genuinely silent.
 const SILENCE_RMS_FLOOR: f32 = 0.012;
+/// Activity `min_rms` for the system lane's turn detection. The pre-transcription
+/// silence pre-filter must not drop below what detection can still find, so both
+/// derive from this one constant.
+pub const SYSTEM_DETECTION_MIN_RMS: f32 = 0.006;
 
 #[derive(Debug, Clone)]
 pub struct DetectionSource {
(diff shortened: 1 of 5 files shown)


andrewwashuta:
@junhohong second set of eyes on this would be good. i've tested it thoroughly

PR #572, opened 2026-07-01
Clay-primary earthy palette, brand-wash surfaces, and icon regeneration pipeline

andrewwashuta:
## What

**Palette.** Clay `#b5551f` becomes the primary brand color: default accent, logo mark, base app icon, and dock icon fallback. The amber preset is removed (it crowded clay); stored `amber` picks legacy-alias to clay, mirroring the existing `blue -> ocean` alias. The remaining presets are retuned richer (oklch chroma ~0.085-0.09) and hue-shifted earthier:

| Preset | Hex | Shift |
|---|---|---|
| Clay (default) | `#b5551f` | terracotta at hue 46, max chroma; distinct from both amber ochre and Claude-style coral |
| Rose | `#a5655c` | richer, same dusty hue |
| Sage | `#527f4d` | warmed toward moss |
| Ocean | `#3d7b9a` | pulled off azure toward storm/petrol |
| Plum | `#965d84` | orchid -> fig/wine |

All presets hold >= 4.5:1 white-text contrast.

**Surface washes.** Each preset now carries a `wash` (the accent with chroma capped at 0.07). Surface neutrals (background, card, borders, sidebar) mix from `--brand-wash` instead of `--brand`, so high-chroma clay tints the greys exactly as quietly as the dusty presets. Buttons and highlights keep the full-strength accent. New `--brand-line` / `--brand-line-strong` tokens give the agent hero chips a clean pastel accent border (opaque oklch mix into the card, not an alpha veil).

**Icon pipeline.** `npm run icons` (`scripts/generate-icons.mjs`) regenerates the full base platform set from `june-app-icon.svg` plus one themed dock icon per preset, parsing `BRAND_PRESETS` from `brand.ts` as the single source of truth and pruning stale icons. The six hand-maintained per-brand SVGs collapsed into one template. Docs in `src-tauri/icons/README.md`.

**Select popovers.** Aligned-open popovers now overlay their trigger pixel-exact in both axes (the offset math missed the popover border, and the swatch track/gap didn't match the trigger's geometry).

**Command palette.** "Open settings" quick action, plus per-tab settings jumps that surface once a query is typed; "appearance"/"accent"/"theme" match the General tab.

## Why

The team wanted the original orange-family brand back as the primary without reading Claude-adjacent, the palette less muted, and surface tinting consistent across accents. The icon pipeline makes future palette changes a one-command regeneration instead of hand-maintained rasters.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #573, opened 2026-07-01
Icon-only chat turn actions with timestamp, scroll-to-latest pill, and branch session fix

andrewwashuta:
## Summary

- Per-turn chat actions (copy, branch) are now icon-only square targets with a muted timestamp on the row's outer side, revealed on hover as before. The visible text labels and the Edit action (which only refilled the composer and doubled context) are removed; copy confirms with a wider checkmark.
- The action row is chromeless and right-aligned under user messages (fixed a silent bug: the inner row was inline-flex, so its auto margin resolved to 0 and the row sat left). Steering rows now sit on the sender's side of the column like user bubbles.
- New floating "Scroll to latest" pill: appears when the reader leaves the live edge (48px threshold shared with stick-to-bottom), jumps back smoothly, and re-engages follow mode without the glide being misread as a user scroll. Anchored inside the fixed composer column so it rides the box height; absolute, never fixed (WKWebView clips composited fixed elements). Styled on the popover surface with the large system shadow, brand-colored arrow, and a light brand wash on hover.
- Turn actions and the timestamp use the shared HoverTip with the deliberate 550ms delay. HoverTip itself now sizes to its content (the width prop is a max): it measures pre-paint, clamps to the viewport, then reveals, and hiding fades out briefly. The compact chip keeps the popover surface with tighter padding.
- Fixed the recurring "Session not found" when branching (root cause below), with a regression test.

## Root cause

`session.branch` targets the live runtime session id, but `branchFromMessage` sent the stored id carried by the turn. That works only while the runtime that minted the id is alive; after an app restart or runtime teardown every branch attempt 404s "Session not found". Branch now tries the cached runtime id and, when that runtime is gone, resumes the stored session and retries once - the same cached-or-resume pattern `fetchSessionUsage` uses.

## Validation

- Full frontend suite green: 128 files, 1883 passed, 2 skipped. New coverage: branch 404 -> resume -> retry; scroll-pill visibility/inertness and CSS pins; HoverTip fit-content, exit fade, and re-entry cancel; a MutationObserver test proving hover reveal cannot move the transcript.
- `pnpm typecheck` and Biome clean on touched files.

- [x] Tested visually where it matters (iterated live with the reviewer through several rounds of visual feedback).

## Out of scope

- Transcript markers/date separators (June already has context-compaction and steering markers).
- Virtualization or content-visibility work for very long transcripts.
- Any change to the stick-to-bottom effects themselves; the pill reuses their conventions.

## Followups

- The HoverTip content-sizing benefits all existing tips (tab bar, privacy cards); no further work planned unless visual QA surfaces edge cases.

- [x] Security-sensitive changes were called out in the summary. (None.)
- [x] Workflow action references are pinned to full-length commit SHAs. (No workflow changes.)
- [ ] Desktop signing, updater, entitlement, capability, or release changes have owner review. (None.)
- [ ] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. (None.)


andrewwashuta:

**Addressed**
- Stale comment in HoverTip's show() → reworded: setPhase("open") guards the interim render before the measure effect lands (a fresh anchor object means the effect always re-runs; there is no skipped case).
- firstElementChild observation assumption in AgentScrollToLatestButton → the ResizeObserver now watches all of the scroller's children, with a comment explaining why child growth (not the fixed-height scroller box) is the signal.

Please take another look.

andrewwashuta:
@greptileai one more small update on top of the earlier fixes: merged main (the clay palette and brand-wash surfaces from #572 landed) and aligned the scroll pill's hover to the new canonical var(--brand-tint) token instead of a hand-rolled brand mix. Please review the latest.

andrewwashuta:
@greptileai final follow-up: applied one Codex finding that was clearly correct — branch runtime resolution now keys off the source session being branched (delegated turns carry their own session id) instead of the selected session, which only picks the gateway. Please confirm the verdict holds on the latest push.

PR #575, opened 2026-07-02
ci: bump Homebrew tap cask on production release

junhohong:
## What this does

- Adds a "Bump Homebrew tap cask" step to `promote-desktop.yml` (the workflow that publishes stable macOS releases; the old `production-desktop-dmg.yml` was retired in #529, so the step lives where the stable DMG is now published). After "Notarize, verify, and publish stable release assets" and before "Record released version on main", it:
  - computes the sha256 of the published `June_universal.dmg`
  - shallow-clones `open-software-network/homebrew-tap` with the release App token
  - rewrites the `version` and `sha256` lines in `Casks/june.rb`, verifying both lines exist beforehand and both rewrites landed afterwards (fails loudly otherwise, so a stale tap is visible)
  - commits as `github-actions[bot]` with message `june v$VERSION` and pushes directly to the tap's `main`
  - is idempotent: a rerun where the cask already holds this version and sha logs and exits 0
- Adds `homebrew-tap` to the release App token's `repositories:` list.
- Documents the Homebrew install path in the README Download section (`brew install --cask open-software-network/tap/june`), keeping the direct DMG link.

## Manual prerequisite

The release GitHub App (the one behind `RELEASE_APP_ID` / `RELEASE_APP_PRIVATE_KEY`) must be installed on `open-software-network/homebrew-tap` with contents write access, or the token mint will fail. This is a one-time org admin action.

## Scope note

This step maintains only the org tap. The in-flight homebrew/homebrew-cask submission is handled separately; once accepted, BrewTestBot autobumps that one.

## Validation

- `actionlint` passes on the workflow.
- The cask rewrite was dry-run locally against a fresh clone of the real `Casks/june.rb`: bump rewrites both lines correctly, rerun with the same version+sha exits 0 with a log line, same version with a different sha rewrites the sha, and a cask missing the expected two-space lines fails before any write.

### DIFF
--- a/.github/workflows/promote-desktop.yml
+++ b/.github/workflows/promote-desktop.yml
@@ -45,7 +45,9 @@ jobs:
           app-id: ${{ secrets.RELEASE_APP_ID }}
           private-key: ${{ secrets.RELEASE_APP_PRIVATE_KEY }}
           owner: open-software-network
-          repositories: os-june,os-june-releases
+          # homebrew-tap: the tap cask bump below pushes as this same release
+          # App, so the App must be installed on that repo too.
+          repositories: os-june,os-june-releases,homebrew-tap
 
       # Read the rc build manifest first (no checkout needed) to learn which
       # version + source commit to promote, then check that commit out below.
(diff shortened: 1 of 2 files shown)


junhohong:
Addressed the Greptile P2: the tap bump now stages only Casks/june.rb explicitly instead of committing with -am. actionlint passes. Marked ready for review.

PR #579, opened 2026-07-02
Move local model config under More options in Settings -> Models

junhohong:
## What

The Models settings tab now shows only the primary model pickers by default. The full local model configuration (enable toggle, endpoint, model id, API key, test connection, remote-endpoint confirm) moved inside the existing "More options" (Advanced model settings) disclosure, alongside the Venice API key it already held.

Judgment calls:
- "More options" auto-expands when a local model is already enabled, so an active setting is never hidden. It only ever expands; a manual collapse stays collapsed.
- Selecting a remote local endpoint from the model picker also expands it, so the "confirm in the Local model section" status message points at something visible.
- Draft endpoint state lives in the parent component, so collapsing never discards in-progress edits.

## Why

Local model configuration is an advanced, uncommon setting; surfacing it as a first-class section crowded the Models tab. Requested placement: under More options (advanced model settings).

## Validation

- `pnpm typecheck` clean; `pnpm test` 127 files / 1902 passed (3 existing local-model tests updated to expand the disclosure first; 2 new tests: collapsed by default + auto-expand when enabled); `pnpm check` (Biome) identical warning count to main, no errors
- Live walkthrough video to follow as a PR comment

## Known gaps

- None known; picker behavior (including the synthetic local option) is unchanged.

### DIFF
--- a/src/components/settings/AppSettings.tsx
+++ b/src/components/settings/AppSettings.tsx
@@ -874,6 +874,9 @@ export function AppSettings({
     const baseUrl = providerSettings.localGeneration.baseUrl.trim();
     if (!isLoopbackUrl(baseUrl)) {
       setLocalEnableConfirm(true);
+      // The confirm affordance lives in the Local model section behind "More
+      // options"; reveal it so the status message's instruction is reachable.
+      setShowMoreModelOptions(true);
       setStatus(
         "This endpoint is not on this machine. Requests will leave your device. Confirm in the Local model section to enable it.",
       );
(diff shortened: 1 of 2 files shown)


junhohong:
QA video artifact

- Video: https://app.opensoftware.co/api/v1/files/fil_67TucJu6npoZ/download
- Compressed file: `20260701-models-more-options.compressed.mp4` (462.2 KiB)

junhohong:
@codex review

PR #333, opened 2026-06-17
Fix meeting live transcript preview

junhohong:
## Summary
- feed live transcript preview from system audio during dual-source meeting recordings by tailing the partial system WAV
- emit live preview events with the correct audio source and source-specific segment ids
- expose `livePreviewEnabled` in recording status so the UI only shows the preview waiting state when the backend actually started preview

## Validation
- `pnpm lint`
- `pnpm test`
- `pnpm build`
- `pnpm format`
- `pnpm test:rust`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked`

## Known gaps
- I did not run a manual live meeting capture in the desktop app from this environment.

PR #580, opened 2026-07-02
Harden system-audio live transcript preview

junhohong:
## Root cause investigation

The orchestrator's prime suspect was that `read_wav_layout` fails to parse the WAV that the macOS system-audio helper writes with `AVAudioFile(forWriting:)`, silently disabling the system live-preview lane. **I empirically disproved that as the cause on current macOS**, then found two real latent defects in the same lane.

### Evidence the WAV path already works on current macOS
Wrote a WAV with the helper's exact settings (`AVAudioFormat(commonFormat: .pcmFormatInt16, ... interleaved: true)`) and hexdumped it:
```
0000: 52 49 46 46 f8 0f 00 00 57 41 56 45 4a 55 4e 4b  RIFF....WAVEJUNK
0030: 66 6d 74 20 10 00 00 00 01 00 02 00 80 bb 00 00  fmt ............
0040: ... 46 4c 4c 52 a8 0f 00 00                       ...FLLR....
0ff8: 64 61 74 61 00 00 00 00                           data....   <- data at 4096, size 0
```
Layout: `JUNK`(28) + `fmt `(PCM, 16-bit, 48k/2ch) + `FLLR`(4008) + `data`(size 0), PCM at offset 4096.
- Ran the actual `read_wav_layout` logic against it -> `data_offset=4096, sample_rate=48000, channels=2` (parses fine; also verified mono, 44.1k, and header-before-first-write).
- Ran the full `WavTailReader` + worker chunking against a real growing AVAudioFile -> emits a proper 768000-sample (8s) chunk.
- Verified a separate process sees AVAudioFile writes in real time while the file stays open (cross-process tail works).
- Frontend: `RecordingSource::System` serializes to `"system"` (camelCase), `upsertLiveTranscriptEvent` never drops system turns (fuzzed 3000 orderings), and `NoteEditor` renders system live turns as `System`. Gating (`configured() && cached_signed_in()`) is identical to the mic lane; helper writes to the same `system.partial.wav` the worker tails.

So on this Mac the lane functions end to end. The two fixes below address the real latent defects that can still keep system audio out of the live transcript.

## What changed

**1. `read_wav_layout`: seek-based chunk walk (was a bounded 64KB scan).**
The old code buffered only the first 64KB and walked chunks in memory. AVAudioFile page-aligns PCM behind `JUNK`/`FLLR` padding; if that padding ever pushes `data` past 64KB, the parser returned `Ok(None)` **forever**, silently disabling the whole system lane (the worker just polls quietly) - exactly the failure mode flagged. Proven: the old scan returns `None (LANE SILENTLY DISABLED)` for a header with data past 64KB, while parsing the normal 4096-aligned header fine. On today's macOS the header is 4KB-aligned so this is latent, not the confirmed repro, but AVAudioFile header layout is macOS-version-dependent, so hardening it removes a zero-diagnostic failure mode. New walk seeks chunk-by-chunk (bounded memory, any offset), preserving the fmt validation and partial-header behavior.

**2. System worker backpressure parity with the mic lane.**
The mic lane drops stale batches (`should_drop_stale_batches`/`newest_preview_batch`) to stay near real time. The system lane, added in the same PR (#333), tails an unbounded file with **no** backpressure, so a slow ASR round-trip lets the backlog grow without bound and the preview lags the live meeting arbitrarily (perceived as current system audio never appearing). Added `trim_stale_system_backlog`: caps the buffer at two chunks, dropping whole stale chunks (chunk-aligned, so timestamps/segment ids stay correct) and advancing `buffer_start_sample`/`segment_index`. Matches ADR 0002's guardrail that preview audio may be dropped to keep up; never touches the recording (saved audio remains the source of truth).

**3. Regression tests.**
- Rust: real AVAudioFile header (`JUNK`+`fmt`+`FLLR`+`data`-size-0) through `read_wav_layout` + `WavTailReader`.
- Rust: `data` chunk past 64KB - **fails on main** (returns `None`), passes here.
- Rust: `trim_stale_system_backlog` whole-chunk trimming + no-op cases.
- Frontend: a system-source live preview turn renders as `System` alongside the mic turn.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml --locked --lib`: 329 passed, 0 failed, 3 ignored (incl. 4 new).
- `cargo fmt --check`: clean.
- `npx vitest run`: 127 files, 1903 passed, 2 skipped, 0 failed (incl. new note-editor test).
- `tsc --noEmit`: clean. (Biome lint not runnable in this env - repo installs deps via bun, which skips the biome platform binary; the added test mirrors surrounding style.)

## Live-walkthrough status
A full in-app meeting-notes walkthrough needs a signed-in account + June API + real system audio, which isn't available here. Per the task's fallback, I verified the WAV-tail path against **real** AVAudioFile-written files end to end (parse -> tail -> 8s chunk emitted) and cross-process real-time visibility - counted as the live evidence for the WAV lane.

## Known gaps / notes for review
- No confirmed in-app reproduction of "system produces nothing" - the happy path works on this Mac. Fixes target the two real latent defects (silent-disable on large headers; unbounded preview lag). If the maintainer can share a failing recording's `system.partial.wav` header, that would confirm whether fix #1 is the exact trigger on their macOS build.
- `validation.rs` has a separate private `read_wav_layout`; only the live-preview one changed.

### DIFF
--- a/src-tauri/src/audio/live_preview.rs
+++ b/src-tauri/src/audio/live_preview.rs
@@ -767,7 +767,7 @@ mod tests {
         // JUNK padding chunk (28 bytes, exactly what AVAudioFile emits).
         buf.extend_from_slice(b"JUNK");
         buf.extend_from_slice(&28_u32.to_le_bytes());
-        buf.extend(std::iter::repeat_n(0_u8, 28));
+        buf.extend(std::iter::repeat(0_u8).take(28));
         // PCM fmt chunk.
         buf.extend_from_slice(b"fmt ");
         buf.extend_from_slice(&16_u32.to_le_bytes());
@@ -780,7 +780,7 @@ mod tests {
         // FLLR page-alignment padding chunk.
         buf.extend_from_slice(b"FLLR");
         buf.extend_from_slice(&(fllr_len as u32).to_le_bytes());
-        buf.extend(std::iter::repeat_n(0_u8, fllr_len));
+        buf.extend(std::iter::repeat(0_u8).take(fllr_len));
         // data chunk with a zero declared size (still recording).
         buf.extend_from_slice(b"data");
         buf.extend_from_slice(&0_u32.to_le_bytes());
--- a/src-tauri/src/audio/live_preview.rs
+++ b/src-tauri/src/audio/live_preview.rs
@@ -402,6 +402,8 @@ fn trim_stale_system_backlog(buffer: &mut Vec<i16>, chunk_samples: usize) -> usi
     // Round the excess up to a whole number of chunks so we never leave a
     // partial chunk that would shift every later frame boundary.
     let excess_chunks = excess.div_ceil(chunk_samples);
+    // The .min is defensive only: with max_len = 2 * chunk_samples the excess
+    // rounded up to whole chunks always fits inside the buffer.
     let drop = excess_chunks
         .saturating_mul(chunk_samples)
         .min(buffer.len());
@@ -541,7 +543,7 @@ fn read_wav_layout(path: &Path) -> std::io::Result<Option<WavLayout>> {
         }
 
         if chunk_id == b"fmt " {
-            if chunk_data_start + 16 > file_len {
+            if chunk_data_start.saturating_add(16) > file_len {
                 return Ok(None);
             }
             let mut fmt = [0_u8; 16];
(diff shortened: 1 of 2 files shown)


junhohong:
Addressed both Greptile comments: the fmt bound check now uses saturating_add to match the chunk-walk style, and the .min on the backlog drop keeps a comment stating it is defensive only (the whole-chunk excess always fits the buffer). Module tests pass (14/14).

junhohong:
@codex review

junhohong:
Codex P2 addressed: the test WAV header builder now uses std::iter::repeat(...).take(...) instead of repeat_n, keeping the file compatible with the declared rust-version 1.80. Module tests pass (14/14).

PR #592, opened 2026-07-02
fix: auto-title agent sessions in sentence case

andrewwashuta:
## Summary

Auto-titled agent sessions now use sentence case (per spec/sentence-case.md) instead of title case. The LLM prompt now explicitly requests sentence case; the fallback replaces `titleCaseSessionTitle` with a simpler `sentenceCaseSessionTitle` that uppercases only the first letter, preserving acronyms and proper nouns as the user typed them.

## Validation

- [x] Frontend tests pass (`pnpm test`): 19/19 passing in hermes-adapter.test.ts
- [x] Type checking passes (`pnpm typecheck`)
- [x] Rust backend tests pass (`cargo test --lib june_api`): 27/27 passing
- [x] Linting passes (`pnpm check`, `cargo fmt --check`)

## Out of scope

Existing sessions retain their stored Title Case titles; this only affects titles generated from now on.

### DIFF
--- a/src-tauri/src/june_api.rs
+++ b/src-tauri/src/june_api.rs
@@ -1632,10 +1632,10 @@ mod tests {
         );
         assert_eq!(
             clean_agent_session_title(
-                "Create a Quarterly Planning Briefing With Follow Up Action Items",
+                "Create a quarterly planning briefing with follow-up action items",
             )
             .as_deref(),
-            Some("Create a Quarterly Planning Briefing With Follow")
+            Some("Create a quarterly planning briefing with follow")
         );
         assert_eq!(clean_agent_session_title("   "), None);
     }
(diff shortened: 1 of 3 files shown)


andrewwashuta:

**Addressed**
- (Greptile) `cleans_agent_session_titles` fed a Title Case string that no longer reflects LLM output → now feeds sentence-cased input, which also guards against future case-normalisation creeping into `clean_agent_session_title`. One deviation from the suggested diff: the expected string is `"Create a quarterly planning briefing with follow"` — the suggestion's `"…follow-up"` is 51 chars, past the 48-char `AGENT_TITLE_MAX_CHARS` truncation.

**Skipped (with reason)**
- (Codex) Enforce sentence case on the LLM output in `clean_agent_session_title` → skipped. Post-hoc case normalisation can't distinguish proper nouns from Title Case words: it would mangle exactly the titles the example cites ("Open GarageBand" → "Open garageband", and "June", "Sarah", "Monday" likewise). The cleaner stays deliberately case-agnostic; the prompt instruction is the reliable lever.

Please take another look.

PR #598, opened 2026-07-02
Fix chat turn-actions overlap, tooltip flip jump, and wrapped tooltip width

andrewwashuta:
## Summary

- The per-turn hover actions row (copy/branch/timestamp) is positioned out of flow in the inter-turn gap, but the gap (16px) was smaller than the row (26px), so the row overlapped the next turn's "Thought" disclosure. The timeline gap now derives from the row's height (`--agent-turn-actions-h` + breathing room = 32px), keeping the large click targets and staying in sync if the button size changes.
- `HoverTip` flipped to the top whenever less than a fixed 200px remained below the anchor (a threshold sized for the largest explainer card), so compact one-line tips near the viewport bottom flipped unnecessarily - and a re-hover on a still-mounted tip recomputed the side, visibly teleporting the open tooltip from bottom to top. The side is now decided from the measured tip height during the existing hidden measuring pass, and a mounted tip keeps its committed side until it fully closes.
- Tooltips that wrap to two lines (e.g. the disabled branch action's "Branching is available once the message is saved") kept their full width cap, leaving lopsided padding. Tips now use balanced line wrapping and, during the measuring pass, tighten their box to the widest rendered line so the edges hug the text. Guarded for jsdom (falls back to the capped width when line rects can't be measured).

## Root cause

Three distinct causes: (1) the inter-turn gap and the out-of-flow actions row were sized independently, and the row outgrew the gap; (2) `HoverTip` chose its side from a fixed space-below constant instead of the tip's real height, and re-computed the side on every `show()` even while the tip was still mounted; (3) the tip box width was only ever capped, never re-fit to the wrapped line boxes.

## Validation

- `pnpm vitest run src/test/hover-tip.test.tsx src/test/agent-turn-actions-css.test.ts src/test/hermes-session-branch.test.tsx` - 25 passed, including new tests: compact tip near the viewport bottom stays on the bottom when it fits; the side does not change on re-entry while mounted; the timeline gap derives from the row height tokens.
- `pnpm typecheck` and `pnpm check` clean (no new warnings).

- [x] Tested visually where it matters (verified in the running app: overlap gone, tooltip no longer jumps).

## Out of scope

- Branching behavior itself (when the branch action is enabled/disabled) - only its tooltip rendering changed here.
- The pre-existing `noStaticElementInteractions` Biome warning on the HoverTip anchor span.

## Followups

- Branching fixes planned as a subsequent story.

- [ ] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review.

### DIFF
--- a/src/components/ui/HoverTip.tsx
+++ b/src/components/ui/HoverTip.tsx
@@ -41,11 +41,26 @@ type TipCoords = {
 function widestLineWidth(element: HTMLElement): number {
   const range = document.createRange();
   range.selectNodeContents(element);
+  // One rect per inline fragment, not per visual line: inline markup (a
+  // <strong>, a link) splits a single line into adjacent rects. Rebuild each
+  // line by grouping fragments that overlap vertically (rects arrive in
+  // document order) and read its width as the group's horizontal span.
   const rects = range.getClientRects();
   let widest = 0;
+  let line: { bottom: number; left: number; right: number } | null = null;
   for (let index = 0; index < rects.length; index += 1) {
-    widest = Math.max(widest, rects[index].width);
+    const rect = rects[index];
+    if (rect.width === 0) continue;
+    if (line && (rect.top + rect.bottom) / 2 < line.bottom) {
+      line.left = Math.min(line.left, rect.left);
+      line.right = Math.max(line.right, rect.right);
+      line.bottom = Math.max(line.bottom, rect.bottom);
+    } else {
+      if (line) widest = Math.max(widest, line.right - line.left);
+      line = { bottom: rect.bottom, left: rect.left, right: rect.right };
+    }
   }
+  if (line) widest = Math.max(widest, line.right - line.left);
   return widest;
 }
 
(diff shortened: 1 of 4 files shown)


andrewwashuta:

**Addressed**
- `widestLineWidth` measured fragment widths, not line widths (Greptile) → fragments are now grouped by vertical overlap and each visual line's width is read as the group's horizontal span, so inline markup no longer over-tightens the box. Added a regression test with a line split into two fragments.

**Deferred**
- Re-measure constrained by the previously applied width on content change (Codex) → will handle in a follow-up pass once this round settles.

Please take another look.

andrewwashuta:
@greptileai I've pushed one more change addressing the secondary review feedback.

**Addressed**
- Re-measure on content change was constrained by the previously applied tightened width (Codex) → the inline width is cleared before measuring so a longer label can grow back, then handed back to React before commit.
- Side could re-decide when the tip content changes while mounted (Codex) → the committed side now lives in a ref for the tip's whole mounted lifetime; the anchor-side plumbing collapsed into it. Added a regression test for a content swap near the viewport bottom.

Please take another look.

PR #600, opened 2026-07-02
Keep sign-in through auth-service outages and unblock first paint on launch

andrewwashuta:
## Summary

- **Don't sign users out on infra outages**: a failed `/auth/refresh` now counts as a definitive session expiry only when the body is a parsed OS Accounts envelope rejection. An unparseable 4xx (an edge/proxy error page during a hosting outage) is now transient — the user stays signed in with a retriable error, matching the existing 5xx handling.
- **First paint no longer blocks on the network**: new `os_accounts_status_local` command derives signed-in state from the keychain alone; `useAccountStatus` clears the loading gate from it immediately, then the full network snapshot fills in details. The launch pale screen drops from up-to-four serial round trips (each with a 30s timeout) to a keychain read.
- **Stale-while-revalidate identity cache**: the last successful user/balance/subscription snapshot is stored alongside the tokens in the existing keychain entry (no new plaintext files). The launch fast-path and transient-failure statuses return it, so launch and outages show the user's real name and last-known balance instead of generic placeholders. Token rotation merges the cache forward under the refresh lock; the write is skipped when unchanged (status runs on every window focus); sign-out deletes the entry, cache included.
- **Auth failure paths now log** (warn-level, `tracing`): refresh classification (HTTP status, parsed-rejection flag, verdict), snapshot failures, and cached signed-in flips. Status/error codes only — never tokens, bodies, or user data.

## Root cause

During an OS Accounts hosting outage, the edge answered `/auth/refresh` with a 4xx error page instead of an API envelope. `refresh_failure_is_transient` treated **any** 4xx as a definitive rejection, so `os_accounts_status` (which runs on every window focus) reported `signed_in: false` and the app replaced whatever view was open with the sign-in gate — while the user's refresh token was actually still valid (the request never reached the auth server, so the rotating token was never consumed). Separately, the app's first paint was gated on that same full network status call, so the outage also manifested as a long blank loading screen before local notes appeared.

## Validation

- `cargo fmt` / `cargo clippy` clean; 36 `os_accounts` unit tests pass, including new coverage: unparseable 4xx → transient, parsed 401 → session expired, parsed 5xx → transient; old-format keychain entries (tokens only) still parse; rotation preserves the cached snapshot; write-if-changed comparison; transient snapshot failure surfaces cached identity, definitive failure does not; wire `TokenPair` decode ignores cache fields.
- `pnpm typecheck` and `pnpm check` clean; account/gate vitest suites pass 42/42, including a new test that the loading gate clears from the local status even when the full status never resolves.
- The snapshot-cache write holds the same refresh lock as token rotation, so it cannot write back a pre-rotation (consumed) token pair.

- [ ] Tested visually where it matters (UI changes: attach a screenshot or recording).

No visual changes by design — the intermediate "signed in, details loading" state renders through pre-existing fallbacks, now normally invisible thanks to the cache. No June API changes; `AccountStatus` wire shape and all existing commands are unchanged (new command is additive).

## Out of scope

- The serial `/me` → `/billing/balance` → `/billing/subscription` fetches in `fetch_snapshot` could be parallelized; left as is since they no longer block first paint.
- The sign-in gate still replaces the whole app when a *definitive* session expiry arrives mid-use; that's intended behavior.

## Followups

- Consider surfacing a small "account temporarily unreachable" indicator when the status is running on cached data during an outage.

- [x] Security-sensitive changes were called out in the summary (auth failure classification, keychain entry now carries a cached identity snapshot next to the tokens; logging excludes tokens/bodies/user data).
- [ ] Workflow action references are pinned to full-length commit SHAs. 

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -257,6 +257,7 @@ pub fn run() {
             providers::set_local_generation_enabled,
             providers::probe_local_generation_endpoint,
             os_accounts::os_accounts_status,
+            os_accounts::os_accounts_status_local,
             os_accounts::os_accounts_login,
             os_accounts::os_accounts_cancel_login,
             os_accounts::os_accounts_logout,
--- a/src-tauri/src/os_accounts.rs
+++ b/src-tauri/src/os_accounts.rs
@@ -132,6 +132,25 @@ fn merge_rotated_tokens(fresh: TokenPair, previous: Option<AccountSnapshot>) -> 
     StoredAccount::new(fresh, previous)
 }
 
+fn access_token_claims(jwt: &str) -> Option<serde_json::Value> {
+    let payload = jwt.split('.').nth(1)?;
+    let decoded = base64::engine::general_purpose::URL_SAFE_NO_PAD
+        .decode(payload)
+        .ok()?;
+    serde_json::from_slice::<serde_json::Value>(&decoded).ok()
+}
+
+fn access_token_subject(jwt: &str) -> Option<String> {
+    access_token_claims(jwt)?
+        .get("sub")
+        .and_then(serde_json::Value::as_str)
+        .map(str::to_string)
+}
+
+fn stored_account_matches_snapshot(stored: &StoredAccount, snapshot: &AccountSnapshot) -> bool {
+    access_token_subject(&stored.pair.access_token).as_deref() == Some(snapshot.user.id.as_str())
+}
+
 #[derive(Deserialize)]
 struct MeWire {
     id: String,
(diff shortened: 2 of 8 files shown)


andrewwashuta:

**Addressed**
- (Greptile) Documented the OS Accounts refresh-envelope contract more explicitly at the refresh failure classifier.
- (Greptile) Detached the best-effort cached snapshot write with `tokio::spawn` so `os_accounts_status` no longer waits on the refresh lock before returning.
- (Codex advisory) Added a token-subject guard before writing a cached snapshot so a stale background task cannot attach one OS Accounts user's identity/balance to another user's token pair after logout/login.

**Verification**
- `cargo fmt`
- `cargo test os_accounts --lib`
- `pnpm test:rust`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features -- -D warnings`

Please take another look.

PR #617, opened 2026-07-03
Fix Hermes routines cron import shadowing

junhohong:
## Summary
- Pin Hermes' core `cron` package from a shared `sitecustomize.py` in bundled runtimes so plugin path insertion cannot shadow it with `plugins/cron`.
- Patch fresh and already-current managed Hermes runtimes by writing the same startup shim before June returns the managed command.
- Add a Hermes smoke gate for `/api/cron/jobs?profile=default`, which is the endpoint the Routines page reads.

## Root cause
Routines calls Hermes' default-profile cron jobs endpoint. In the packaged dashboard, plugin discovery can prepend `hermes-agent/plugins` to `sys.path`, so `from cron import jobs` in Hermes' dashboard route resolves to `plugins/cron` instead of core `hermes-agent/cron/jobs.py`. That raises an ImportError and surfaces as the 500 shown on the Routines page.

## Validation
- `python3 -m py_compile src-tauri/src/hermes/sitecustomize.py`
- `bash -n scripts/bundle-hermes-runtime.sh`
- `pnpm exec biome check scripts/hermes-smoke.ts`
- `pnpm exec vitest run src/test/hermes-smoke.test.ts src/test/hermes-routines.test.ts`
- `pnpm run typecheck`
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked hermes_bridge -- --nocapture`
- Temporary-shim runtime smoke with the installed Hermes binary: `PYTHONPATH=<temp sitecustomize> JUNE_HERMES_COMMAND=/Applications/June.app/Contents/Resources/native/hermes/bin/hermes pnpm test:hermes-smoke`, including `PASS /api/cron/jobs?profile=default returned 0 job(s)`.

## Not run
- Windows PowerShell bundle script execution. PowerShell is not installed on this macOS host.


junhohong:
Follow-up on review feedback addressed in latest commit `eba991c0`:

- The shim now has a single source of truth: `src-tauri/src/hermes/sitecustomize.py`, copied by bundle scripts and embedded by Rust with `include_str!`.
- `hermes_runtime_available_for_auto_start` is pure again. Managed-runtime shim installation happens in `resolve_hermes_command` before returning the managed command.
- Managed runtime shim installation now writes to every discovered Unix `python*/site-packages` directory instead of choosing the lexicographically first one.
- Bundle self-tests pass the plugin root via `HERMES_PLUGIN_ROOT` instead of interpolating it into the Python command string.

Final CI is green on `eba991c0`.

PR #620, opened 2026-07-03
Validate saved Venice API keys

junhohong:
## Summary
- require newly saved and user-supplied Venice API keys to use the `VENICE_INFERENCE_KEY_` prefix
- verify saved Venice keys against Venice `/models` before persisting them using a pooled Tauri HTTP client
- preserve legacy saved keys so existing users get an explicit `venice_api_key_invalid` message instead of a silent fallback
- return `venice_api_key_invalid` / `venice_api_key_rejected` instead of generic upstream failures when user-provided keys are malformed or rejected

## Tests
- `cargo test --manifest-path src-tauri/Cargo.toml venice_api_key --lib`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --all-features --locked -- -D warnings`
- `cargo clippy --all-targets --all-features --locked -- -D warnings` (from `june-api/`)
- `cargo test -p june-api --all-features --locked`
- `cargo test -p june-providers --all-features --locked -- --test-threads=1`

Note: an unconstrained local `cargo test -p june-providers --all-features --locked` hit OS file descriptor exhaustion from concurrent wiremock/JWKS tests; the same package passed serially.

### DIFF
--- a/june-api/crates/api/src/auth.rs
+++ b/june-api/crates/api/src/auth.rs
@@ -3,6 +3,7 @@ use axum::http::{HeaderMap, header};
 use june_domain::{ProviderCredentials, UserId};
 
 const VENICE_API_KEY_HEADER: &str = "x-venice-api-key";
+const VENICE_API_KEY_PREFIX: &str = "VENICE_INFERENCE_KEY_";
 
 pub(crate) async fn authenticated_user(
     state: &ApiState,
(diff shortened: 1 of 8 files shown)


junhohong:
Follow-up on the Greptile review for older commit `e2b018c`:

- centralized `user_venice_key_auth_error` in `venice.rs` and reused it from augment/image/image edit
- preserved legacy saved Venice keys during settings sanitization so users get the explicit `venice_api_key_invalid` path instead of a silent drop
- switched Venice key verification to a pooled static Tauri HTTP client

Latest commit `13fdd731` has GitHub checks green and Octopus 0 findings.

PR #637, opened 2026-07-06
Allow user-directed June personas

junhohong:
Closes JUN-211

## Root cause
June's managed Hermes `SOUL.md` identity rule did not distinguish real identity/provenance questions from user-requested role or persona framing, so the agent could treat role requests as identity-reset attempts and refuse them too broadly. While testing the same flow, the Settings image model picker also showed a stale four-model image catalog with no privacy or uncensored model metadata. One more gap: the image tool honored the selected model and safe-mode setting, but the chat model could still pre-refuse image requests in prose before calling the tool.

## What changed
- Updated June's managed Hermes `SOUL.md` seed so explicit user-requested roles and personas are allowed as task framing.
- Kept June's real identity/provenance rule intact: June should still be transparent that it is June when asked what it is.
- Scoped the persona exception to system/developer instructions, privacy, tool limits, the user's own stated constraints, and current app/tool settings, so the prompt does not add a separate safety veto over Settings-controlled behavior.
- Taught the image-tool prompt not to add a separate content refusal layer before calling `june_image`; the selected image model and safe-mode setting are authoritative, and provider rejection is reported only if the provider rejects.
- Expanded the curated image model picker to the current priced Venice image catalog.
- Added visible image-model tags for `Private`, `Anonymized`, and `Uncensored` where Venice metadata provides those signals.
- Synced June API `image_pricing` with the expanded picker list so newly exposed models are accepted by the backend.
- Added regression coverage for the SOUL sync test, image-tool pre-refusal guidance, the image model picker, uncensored search, and backend image pricing defaults.

## Why
The user's explicit instruction is now the control point for role/persona behavior, while identity transparency and app/tool settings still take precedence. The image model picker now exposes all priced models June can generate with and makes the privacy/uncensored tradeoffs visible before selection. Image requests now flow through the configured image provider instead of being blocked by an extra chat-model refusal.

## Validation
- `cargo test --manifest-path src-tauri/Cargo.toml sync_june_soul --locked`
- `cargo test --manifest-path src-tauri/Cargo.toml soul --locked`
- `make tauri-fmt-check`
- `make tauri-test`
- `pnpm test -- src/test/app-settings.test.tsx`
- `cargo test -p june-config default_config_prices_the_curated_image_models --locked`
- `cargo fmt --all --check` from `june-api/`
- `pnpm typecheck`
- `git diff --check`
- GitHub Actions: pending rerun on latest commit `6317f7a0`

## Visual testing
`pnpm tauri:dev` is running locally on the PR branch with local-dev auth aligned. Vite is serving at `http://127.0.0.1:1421/`, June API is listening on `127.0.0.1:8080`, Hermes/system-audio initialized successfully, and the local logs show `image_generate` requests using the selected `lustify-v8` image model.

## June API deploy
June API deploy is now needed before promoting a desktop build that exposes the expanded image list in production. Without the backend `image_pricing` update, newly selectable image models would fail with `model_not_priced`.

## Out of scope
- New Settings toggle for persona behavior. JUN-211 allowed the behavior to be user-controlled through explicit user instruction.
- Live upstream generation tests for every new image model, to avoid spending provider credits during PR validation. The catalog and pricing metadata were verified against Venice's current `/models?type=image` response.

### DIFF
--- a/docs/adr/0008-image-generation-and-editing-tools.md
+++ b/docs/adr/0008-image-generation-and-editing-tools.md
@@ -41,6 +41,15 @@ Phase C has landed. The current built-in fallback image generation price list is
 image. Image editing is metered separately through `image_edit_pricing`, with
 `firered-image-edit` priced at 80 credits per image edit.
 
+## Addendum - 2026-07-06
+
+The Settings image-model picker now mirrors Venice's current image catalog for
+models June can price, including privacy tags (`Private` or `Anonymized`) and an
+`Uncensored` tag when Venice marks the model that way. `image_pricing` remains
+the backend allowlist for generated image models; newly exposed models are priced
+with the same flat ~2x convention as the original list. Models that Venice prices
+by resolution use the default 1K tier until June exposes resolution controls.
+
 ## Context
 
 JUN-129 shipped `/image <prompt>` as a client-side slash command: it creates a
--- a/june-api/crates/config/src/lib.rs
+++ b/june-api/crates/config/src/lib.rs
@@ -593,15 +593,45 @@ fn default_pricing() -> BTreeMap<String, ModelPriceConfig> {
 /// and `DEFAULT_IMAGE_MODEL` in the Tauri providers module — every id here must
 /// be a current Venice image model (verified against the models list), or
 /// generation fails `image_generation_rejected`. Values are the Venice per-image
-/// cost with a ~2x margin (mirroring the flat web-tool pricing): SD3.5 ~$0.01 ->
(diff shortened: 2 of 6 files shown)


junhohong:
Greptile ADR suggestion triage: not taking for this PR. The repo's ADR rule requires all three criteria, including a hard-to-reverse decision such as an architectural shape, integration contract, or boundary. This change is a managed prompt seed update that can be revised or replaced by a future Settings toggle without migration or contract cost, so I documented the root cause, out-of-scope toggle, and rationale in the PR body instead.

junhohong:
@codex review

PR #639, opened 2026-07-06
Do not authorize or charge users for issue reports (JUN-213)

andrewwashuta:
Closes JUN-213

## Summary

- Issue reports (bug / feedback / feature) no longer run a model turn, so nothing is authorized or charged — fixed by construction. Reports move from composer category chips to a direct-submit popover: composer "+" rows, the sidebar, and settings "Report an issue" all open it (category preselected); it collects a description and drag-and-drop / picked file attachments and submits straight to `/v1/issue-reports` with no `sessionId` and no `agentDiagnosis`.
- The team-facing diagnosis moves into June API: when a report arrives without a client diagnosis and `issue_reports.diagnosis_model` is configured, June API runs one non-streaming completion on that model at its own expense — never metered, capped by `diagnosis_timeout_secs` (default 30) and `max_tokens: 1024`, and any failure delivers the report undiagnosed. Output feeds the existing diagnosis splitter; heading-less prose degrades gracefully to a single combined issue.
- Report categories are removed from the "/" slash menu. The chip machinery stays wired (restored drafts with old chips still submit as before); only the entry points that insert new chips are gone.
- ADR 0012 records the decision (including why the client waiver-flag + server model-pin design was rejected); CONTEXT.md gains the canonical "Issue report" term.
- Drive-by fix the QA walkthrough caught: `.agent-attach-menu` never opted back into pointer events inside the `pointer-events: none` docked composer, so every "+" menu row (including "Attach files") was click-dead by mouse outside the hero composer. Same one-line pattern the model popover documents.

## Root cause

Three frontend comments described a server-side "no-charge waiver" keyed off the report category, but no waiver was ever implemented in June API or OS Accounts. The investigation turn a category chip triggered was metered and charged as ordinary agent chat at whatever generation model the user had selected, so users paid to file reports and the price scaled with their model choice.

## Validation

- `make verify` green (Biome, typecheck, vitest 132 files / 2105+ passed, cargo fmt/clippy/test for both Rust crates), re-run after rebasing onto current main.
- New coverage: `report-popover.test.tsx` (payload shape without session fields, attachments-only fallback, drop routing, submit blocked while a drop import is in flight, failure keeps input), june-api `issue_reports` service tests (diagnosis added / passed through / skipped / timeout / provider error, config validation), plus updated "+"-menu, slash-menu, and seeded-category tests.
[...]

PR #641, opened 2026-07-07
Paste screenshots into the issue report dialog

andrewwashuta:
## Summary

- Pasting an image (a CleanShot screenshot, for example) into the issue report dialog attaches it through the same import pipeline as drag and drop; text pastes fall through to the textarea untouched.
- The composer's clipboard image extraction moves to `src/lib/clipboard-files.ts` so both surfaces share one implementation (WKWebView delivers pasted screenshots as Files; Finder file copies never reach `clipboardData`, so image-only interception matches what can actually arrive).
- Drag-over now reads as a real drop zone: a gray dashed panel washes over the form with a centered "Drop files to attach" instruction, with drag depth counted so child edge-crossings do not flicker the overlay.

Follow-up to #639 (JUN-213) — both requested during design review, landed just after the merge.

## Validation

- `pnpm typecheck`, Biome, and the report-dialog + agent-workspace vitest suites green (new tests cover pasted-image routing and text-paste passthrough).
- Drop-zone state verified visually in the browser harness (drag-over screenshot reviewed).
- [ ] Needs a June API (backend) deploy: no.

## Out of scope

- Pasting non-image files (they never materialize in WKWebView clipboardData).
- A permanently visible dropzone box in the resting form (the ghost Add files button, paste, and drag-over cover the affordance without the vertical cost).

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -230,6 +230,7 @@ import {
   isTopUpRequiresMaxError,
   messageFromError,
 } from "../../lib/errors";
+import { clipboardImageFiles } from "../../lib/clipboard-files";
 import { withTimeout } from "../../lib/async-timeout";
 import {
   MESSAGING_PLATFORMS_LOAD_TIMEOUT_MESSAGE,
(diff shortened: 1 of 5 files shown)


andrewwashuta:

**Addressed**
- Mixed paste drops text (P1) → an image-carrying paste that also carries meaningful text now imports the image AND lets the text insert normally; only image-only pastes prevent default (keeping screenshot-tool metadata out of the field). New test covers the mixed case.
- Stop dialog pastes from bubbling into the composer (Codex P2) → already fixed in 05b59a5e (pushed shortly after the review): the dialog stops propagation on its drag/paste events and the composer importers ignore events while the dialog is open, with a regression test.

**Skipped (with reason)**
- Partial items hide files (P2) → the items-first extraction is the composer's long-shipped behavior in the shared helper; unioning items with files risks double-attaching the common case (browsers mirror the same entries in both) to serve a payload shape we have not observed. Happy to revisit with a concrete repro.
- Duplicate names collapse attachments (P2) → the collapse is the helper's deliberate multi-representation handling (the same screenshot arriving as png+tiff picks the best rank); treating duplicate-name payloads as distinct images would re-break that common case. Pre-existing, intentional trade-off, not introduced by this PR.

Please take another look.

PR #642, opened 2026-07-07
Adopt shadcn shimmer recipe across text and HUD shimmers

andrewwashuta:
## What

Replaces June's four hand-rolled text-shimmer implementations with a single vendored plain-CSS port of shadcn's `shimmer` utility, and reshapes the three HUD mark shimmers to match the same band geometry. One shimmer look across the whole app, net fewer lines.

- **New `src/styles/shimmer.css`** — faithful port of the `shimmer` utility from the `shadcn` package (v4.13.0), hand-written for June (which has no Tailwind). Keeps shadcn's variable API verbatim (`--shimmer-duration`, `--shimmer-spread`, `--shimmer-angle`, `--shimmer-color`, ...) so upstream docs apply. The band is currentColor-based with a soft wide highlight, font-relative spread (`calc(3ch + 40px)`), and a 20deg tilt; it brightens automatically in dark mode via `[data-theme="dark"]`.
- **Text sites migrated** — `.text-shimmer` (agent chat "Thinking…", image labels, relaunch title), `.transcript-processing-label` (recorder), `.note-processing-roll-item` (rolling stage names), and the notes-list `[data-shimmer]` status word now share the utility. Each site keeps its own `color` and cadence (agent 1.2s, notes list 1.5s, others 2s). Old `note-text-shimmer` / `note-status-shimmer` keyframes deleted.
- **Skeleton bars** — the usage-panel first-load bars are a block background sweep (not text-clip), so they got their own `usage-skeleton-sweep` keyframe rather than riding a text-shimmer keyframe.
- **HUD marks** — the agent, meeting, and dictation HUD logo shimmers (a highlight swept across a masked glyph) were reshaped to the shadcn band: single peak at 50%, soft 50%-mix shoulders, tilt carried by the gradient angle (`skewX` removed). Cadence (1320ms), accent colors, mask setup, and paused/reduced-motion behavior are unchanged, so the "live" signal reads the same strength.

## Why

Several independently-tuned gradients with hard 35/50/65 stops had drifted; adopting a design-system recipe unifies the look, makes future tuning a one-variable change, and tracks upstream. The `shimmer` semantics shift to currentColor-based (the element carries a real `color`, no more `color: transparent`).

## Design notes

- **Highlight polarity:** shadcn's default *fades* the text toward the surface (currentColor at 20% alpha). We kept that default; if the small muted status labels read washed out in practice, `--shimmer-color: var(--foreground)` flips it to the old *intensify* direction — a one-line escape hatch, not a fork.
- **Spread is font-relative, not width-relative** — deliberately chosen so short status labels all carry the same physical glint (one light source across the app) rather than each label sizing its band to its text length.

## Tested visually

Needs a look in-app: agent chat "Thinking…", recorder transcribing, notes-list status word, and the three HUDs (`__agentHud` / `__meetingHud` / `__dictationHud` console commands park the states). No screenshots attached yet.

## Needs a June API deploy?

No — desktop-only CSS/markup change.

## Out of scope

- The JS-driven audio-meter carrier shimmer (`src/hud.ts`) — a different mechanism, untouched.
- The `shadcn` npm package is **not** a dependency; `shimmer.css` is the vendored copy (`pnpm dlx shadcn` fetches upstream to diff a newer version).

## Followups

- If we ever shimmer a long line (e.g. a skeleton paragraph), a percentage `--shimmer-spread` at that one call site gives width-relative behavior deliberately.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/642"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785994575&installation_id=144203540&pr_number=642&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F642&signature=1185189b05780d7f46625db77a78f1ccae7b47e3c1b253b69912993e6d94a99f"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -3429,7 +3429,7 @@ function UpdateRelaunchCard({
         </span>
         <span className="update-relaunch-copy">
           <span
-            className={relaunching ? "update-relaunch-title text-shimmer" : "update-relaunch-title"}
+            className={relaunching ? "update-relaunch-title shimmer" : "update-relaunch-title"}
           >
             {relaunching ? "Relaunching..." : "Relaunch to update"}
           </span>
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -10294,7 +10294,7 @@ function AgentChatTurnRow({
         />
         {textParts.length === 0 && nonTextParts.length === 0 ? (
           <p className="agent-assistant-empty">
-            <span className="text-shimmer">Thinking…</span>
+            <span className="text-shimmer shimmer">Thinking…</span>
           </p>
         ) : (
           // No actions on an empty/in-flight turn. There is nothing useful to
@@ -10703,7 +10703,7 @@ function AgentGeneratedImage({
     return (
       <div className="agent-generated-image" data-status="running" role="status" aria-live="polite">
         <div className="agent-generated-image-placeholder">
-          <span className="text-shimmer">Generating image…</span>
+          <span className="text-shimmer shimmer">Generating image…</span>
         </div>
       </div>
     );
@@ -10730,7 +10730,7 @@ function AgentGeneratedImage({
   const image = imageSrc ? (
     <img src={imageSrc} alt={part.prompt} draggable={false} />
   ) : part.path ? (
-    <span className="agent-generated-image-loading text-shimmer">Loading image...</span>
+    <span className="agent-generated-image-loading text-shimmer shimmer">Loading image...</span>
   ) : null;
   return (
     <figure className="agent-generated-image" data-status="complete">
@@ -11468,7 +11468,7 @@ function AgentThinkingGroup({
       onToggle={(event) => onOpenChange(event.currentTarget.open)}
     >
       <summary>
-        <span className={running ? "text-shimmer" : undefined}>
+        <span className={running ? "text-shimmer shimmer" : undefined}>
           {running ? "Thinking" : "Thought"}
         </span>
         <IconChevronDownSmall size={14} className="agent-disclosure-chevron" />
@@ -12904,14 +12904,14 @@ function ActivityIndicator({
   );
 }
 
-// Bottom-of-timeline "responding" affordance: a shimmering label, reusing the
-// same text-shimmer the recorder uses while transcribing. Lives in the timeline
-// (not the header) so it reads like the agent is actively composing the next
-// turn.
+// Bottom-of-timeline "responding" affordance: a shimmering label, painted by
+// the same shared .shimmer utility the recorder uses while transcribing. Lives
+// in the timeline (not the header) so it reads like the agent is actively
+// composing the next turn.
 function AgentThinking() {
   return (
     <div className="agent-thinking" role="status" aria-live="polite">
-      <span className="text-shimmer agent-thinking-label">Thinking…</span>
+      <span className="text-shimmer shimmer agent-thinking-label">Thinking…</span>
     </div>
   );
 }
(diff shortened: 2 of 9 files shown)


andrewwashuta:
Pushed fixes addressing the review.

**Addressed**
- **Unsupported highlight hides text** (Greptile `shimmer.css:47`, Octopus `shimmer.css:55`) → The text-clipping paint (`background-clip: text`, transparent `-webkit-text-fill-color`, and the `oklch(from ...)` / `color-mix()` gradient) is now gated behind `@supports (color: oklch(from red l c h))`. Engines that reject relative color syntax skip the paint entirely and fall back to the element's solid `color` — legible, just no sweep — instead of rendering the label invisible. This is the graceful fallback the shadcn docs prescribe.
- **Reduced motion still animates** (Greptile `agent-hud.css:296`) → Added `.agent-hud-mark::before` to the reduced-motion `animation: none` list. The sweep lives on the pseudo-element, so it now stops for reduced-motion users, matching the meeting and dictation HUDs.
- **`data-shimmer` split contract** (Octopus, advisory) → Sharpened the `.folder-note-subtitle [data-shimmer]` comment to state the class (paint) and attribute (duration) are both required, so a future refactor doesn't drop the attribute.

**Skipped (with reason)**
- **Fold `.shimmer` into `.text-shimmer`** (Octopus, advisory) → Octopus itself notes this is a style concern, not a bug. Keeping `.shimmer` as a standalone paint utility is deliberate: it lets the `@supports` fallback live in one place and keeps layout/truncation separable from the sweep. Neither half is harmful alone (shimmer without truncation, or truncation without shimmer are both valid).
- **`calc(90deg + 20deg)` → `110deg`** (Octopus, advisory) → Kept intentionally; the `calc()` documents that the HUD tilt tracks `shimmer.css`'s `--shimmer-angle` default (20deg), which a bare `110deg` would obscure. The adjacent comment states this.

Note: the failing **Biome check** is a pre-existing `noConfusingVoidType` error in `AgentActivityDrawer.tsx`, a file this PR does not touch (byte-identical to `main`). It's unrelated to the shimmer work and Biome flags its fix as unsafe, so I've left it out of this PR.

@greptileai please take another look.

andrewwashuta:
Pushed round-2 fixes — both P1s were correct catches, my first pass was incomplete.

**Addressed (Greptile)**
- **Partial support hides text** (`shimmer.css:63`) → Added an inline fallback to the angle: `calc(90deg + var(--shimmer-angle, 20deg))`. Now an engine that passes the relative-color `@supports` gate but doesn't honor the `@property` initial value keeps a valid gradient instead of an invalid one that would leave the label transparent. `--shimmer-angle` was the only gradient input relying on `@property`; the rest already carry inline fallbacks.
- **Reduced motion still wins** (`agent-hud.css:545`) → The sweep and breathe are started by status-qualified selectors (`.agent-hud-mark[data-status="running"]::before`, `.agent-hud-status[data-status="waitingForUser"]`), which out-specify a plain `::before` reset. The reduced-motion block now matches that `[data-status]` specificity, and since it sits later in the file, equal specificity resolves in its favor. Added a comment explaining why the attribute qualifier is required. (The meeting and dictation HUDs don't gate their mark animation on `[data-status]`, so their existing resets are already sufficient.)

@greptileai please take another look.

PR #644, opened 2026-07-07
Extract a design system: docs, specs, living styleguide, and a token sweep

andrewwashuta:
## Summary

Extrapolates a design system from the existing UI and makes it visible, so developer-injected inconsistency has a single source of truth to converge on. First pass is mostly additive plus a low-risk consistency sweep; a broader app-wide sweep is deliberately deferred.

- **Written system** — `docs/design/{foundations,components,conventions,taste}.md`: the theming model, token roles, the type system (scale, heading mapping, two-weight rule, family roles), the pattern-to-canonical-component map with the settings markup contract, the flat-namespace and interaction rules, and a portable taste statement. Each doc links the styleguide and ends with an honest known-deviations list.
- **Five new enforceable specs** — `type-scale`, `font-weights`, `font-families`, `no-all-caps`, `control-sizes`; registered in `spec/index.md`, `docs/index.md`, and the `AGENTS.md` spec list.
- **Living styleguide** (dev-only) — `styleguide.html` + `src/styleguide/`, opened in a plain browser via `pnpm dev` at `http://localhost:1421/styleguide.html`. Renders the real tokens and components with a live theme + brand-preset switcher; foundations sections self-update from `tokens.css` via a `?raw` parse. Never bundled into production (not a Vite rollup input, same pattern as the onboarding preview).
- **Token and type sweeps** — added `--fs-2xs` and `--on-solid`; normalized font weights so declarations match what actually renders (Diatype ships only 400/600 with `font-synthesis: none`, so `500` was silently rendering `400`); swept exact-match raw font sizes and control heights onto tokens; pruned tabular numerals from static numbers, keeping only the live recording timer.
- **Shared `useDismiss` hook** — the same close-on-outside-press-or-Escape block was re-implemented in five menus and popovers; consolidated into `src/lib/use-dismiss.ts`, behavior preserved per site.

## Validation

- Full frontend gate green: `pnpm typecheck` clean, 2112 vitest tests passing.
- Re-verified spec invariants on the merged CSS (only `400`/`--fw-medium` weights, one intentional tabular-nums, no uppercase).
- Production build confirmed to contain zero styleguide assets.
- Ran the repo review battery (Standards / Spec / Adversarial, both harnesses) before opening: clean.

- [x] Tested visually where it matters. The two deliberate visual changes to eyeball: note-preview headings drop 15px to 14px (matching markdown headings), and eight structural titles move to medium weight. Everything else is zero-visual-change by construction.
- [ ] Needs a June API (backend) deploy. No — frontend and docs only.

## Out of scope

Deferred to a later consistency pass (seeded by the known-deviations list in `foundations.md`): a shared `Field`/input treatment and toggle-wrapper renames (they collide with in-flight settings work), a shared `usePopoverPosition` primitive, a `Button` React component, a dark-surface token family for the recording-indicator pill, the lone 15px tool-disclosure glyph, and the borderline medium-weight candidates.

## Followups

- The pass-2 consistency sweep, tracked in the `foundations.md` known-deviations list.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/644"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785997205&installation_id=144203540&pr_number=644&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F644&signature=060c3b3d94713bd134dc6e07cff72ccfa07679ab1ce596cac8f0967d451350a8"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


andrewwashuta:

**Addressed**
- Relative timestamp jitters (`.agent-session-time`) → restored `font-variant-numeric: tabular-nums`. You're right: `Sidebar` runs a 60s clock so this relative time advances and its width would shift on proportional digits. This is exactly the live-value exception in `spec/no-tabular-numerals.md`; the pass-2 audit had wrongly treated it as a static timestamp. Added a comment citing the exception. The activity-drawer ages have no steady clock, so they stay proportional.

**Note (pre-existing, not from this PR):** the Biome CI check is red due to a single non-file-attributable error that landed on `main` with #642 (the check went from green to red at that commit) and is inherited here by rebase. Every file this PR touches is individually Biome-clean. Flagging so it isn't mistaken for a regression from this change.

Please take another look.

andrewwashuta:
@greptileai Pushed two follow-ups after your approval:

- **Fixed the red Biome check.** Root cause found: it was a formatter error, not a lint rule. `src/app/App.tsx` was committed unformatted in #642 (the relaunch shimmer className), which flipped Biome red on `main` and here by rebase. Ran the formatter, whitespace-only, no semantic change. CI Biome is green now.
- **Documented the shimmer utility** in `docs/design/components.md`, since #642 made it a system pattern (pair `.shimmer` with a semantic class, currentColor base, graceful `@supports` + reduced-motion fallback).

No product-code changes beyond the earlier tabular-nums fix. Please confirm the verdict still holds.

PR #645, opened 2026-07-07
Harden Windows Artifact Signing workflow

wayanjimmy:
## Summary

- clean up Azure Artifact Signing metadata files in a `finally` block
- write signing metadata under `RUNNER_TEMP` when running in GitHub Actions
- pin the NuGet/signing toolchain versions and verify the downloaded `nuget.exe` checksum
- make Windows signing secret validation errors distinguish no credentials, partial config, mixed config, and both modes configured

## Root cause

The Azure Artifact Signing path was structurally correct, but the release workflow still had production-hardening gaps: metadata files could remain on disk after a failed signing attempt, the signing toolchain could drift because it used `latest`/unpinned package installs, and setup failures were collapsed into one generic signing-configuration error.

## References

- GitHub Actions variables reference: `RUNNER_TEMP` is the path to a temporary directory on the runner, emptied at the beginning and end of each job when permissions allow: https://docs.github.com/en/actions/reference/variables-reference#default-environment-variables
- GitHub-hosted runner reference: runner file system locations and workflow directory behavior: https://docs.github.com/en/actions/reference/github-hosted-runners-reference#file-systems
- Microsoft Artifact Signing / Trusted Signing SignTool integration uses a metadata file passed through `/dmdf`: https://learn.microsoft.com/en-us/azure/trusted-signing/how-to-signing-integrations

## Testing

- `git diff --check`
- Parsed `.github/workflows/production-desktop-windows.yml` with PyYAML
- Parsed `scripts/windows-sign.ps1` with the PowerShell parser
- Local PFX signing succeeded with a throwaway self-signed certificate
- Local fake Azure Artifact Signing failed as expected and left 0 metadata files in `RUNNER_TEMP`

`actionlint` was not available locally.

## Visual testing

Not tested visually. This is release workflow/script-only.

## June API deploy

No June API deploy needed.

## Out of scope

- proving real Azure Artifact Signing end to end with production credentials
- changing the signing mode decision between PFX and Azure
- renaming existing Artifact Signing secrets

## Followups

The Azure signing path still needs a real Windows production workflow run with the actual Azure signing account, certificate profile, OIDC federation, and production environment secrets before it can be considered fully proven.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/645"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785998763&installation_id=144203540&pr_number=645&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F645&signature=307b6f26058feae55ed7da4f03e16c9f850f9e2656828178f2e925386a8f90c5"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/.github/workflows/production-desktop-windows.yml
+++ b/.github/workflows/production-desktop-windows.yml
@@ -107,13 +107,22 @@ jobs:
           choco install nsis 7zip -y --no-progress
           "C:\Program Files (x86)\NSIS" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append
 
+          $nugetVersion = "6.14.0"
+          $nugetExpectedSha256 = "92DBED160DDEE0F64B901E907439E021211B428E57C089ECC12FC38DCC4BD9A5"
+          $windowsSdkBuildToolsVersion = "10.0.28000.2270"
+          $artifactSigningClientVersion = "1.0.128"
+
           $nuget = Join-Path $env:RUNNER_TEMP "nuget.exe"
-          Invoke-WebRequest -Uri "https://dist.nuget.org/win-x86-commandline/latest/nuget.exe" -OutFile $nuget
+          Invoke-WebRequest -Uri "https://dist.nuget.org/win-x86-commandline/v$nugetVersion/nuget.exe" -OutFile $nuget
+          $nugetActualSha256 = (Get-FileHash -LiteralPath $nuget -Algorithm SHA256).Hash.ToUpperInvariant()
+          if ($nugetActualSha256 -ne $nugetExpectedSha256) {
+            throw "nuget.exe checksum mismatch. Expected $nugetExpectedSha256, got $nugetActualSha256."
+          }
 
           $sdkRoot = Join-Path $env:RUNNER_TEMP "windows-sdk-build-tools"
-          & $nuget install Microsoft.Windows.SDK.BuildTools -x -OutputDirectory $sdkRoot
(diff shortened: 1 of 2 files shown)


wayanjimmy:
Addressed the signing secret presence disclosure in 683585e9 by removing the per-secret `is set` / `is not set` notices from the invalid signing-config path. The workflow still reports the actionable invalid state and, for partial single-mode configs, the missing variables needed to complete that mode.

I also rechecked the Octopus SHA-256 note: the pinned NuGet SHA-256 constant is 64 hex characters and matches a fresh local download of https://dist.nuget.org/win-x86-commandline/v6.14.0/nuget.exe via `Get-FileHash -Algorithm SHA256`, so I left it unchanged.

Validation after the follow-up:
- `git diff --check`
- PyYAML parse of `.github/workflows/production-desktop-windows.yml`

jakubswierczek (approved):

PR #646, opened 2026-07-07
Ask June note chat panel, note toolbar, and dictation-during-recording

andrewwashuta:
## Summary

Rebuilds the "Ask June" experience on meeting notes into a contextual chat panel, adds a persistent note toolbar, and lets dictation run safely while a meeting records.

- **Ask June flies out a note-contextual chat panel** — same attach mechanics as the agent files panel (fixed right card, main card slides via `:has()`, spring entrance, drag-resize). It's a **real per-note Hermes session** (remembered per note, appears in agent history, "Open in agent view" resumes it), with the shared model picker, file attach, dictation, note-tailored presets, and streaming replies. Reuses lib-level plumbing (gateway/event-classifier/turn-builder/composer) rather than touching the AgentWorkspace monolith.
- **Session state is owned by App**, so it keeps running when the panel closes: a fired-off question shows a pulsing **working dot** on the toolbar button until the reply lands, then you reopen to the answer.
- **Persistent note toolbar on every note** — consistent vertical rhythm regardless of navigation origin. Breadcrumb nav on the left, **Ask June + copy-link + an overflow menu (Delete note, confirmed)** on the right. Note metadata (date · project) moved **below** the title, above the Notes/Transcription toggle.
- **Dictation coexists with a live recording** — while dictation listens, the recording's mic channel is ducked (writes silence, not dropped frames, to keep the mic/system tracks time-aligned), so a dictated question never contaminates the meeting note. Rust-only, driven by dictation helper events, restored on every terminal/crash path.
- **HUD waveform hardened** — coalesces bursty `audio_level` events to one per animation frame so the travelling wave can't lurch under CPU contention. Default dictation cadence (≤50Hz < 60Hz frame) is unchanged.
- **Refactor**: extracted `MarkdownContent` and the composer `ModelPicker` out of the 13k-line `AgentWorkspace` so both chat surfaces share them (net −900 lines there).

## Validation

- `make verify`-equivalent green: `pnpm check` (0 errors), `pnpm typecheck`, `pnpm test` (2103 passing), and `cargo fmt --check` / `cargo clippy -D warnings` / `cargo test` for `src-tauri`. `june-api` is untouched.
- New tests: per-note session map, `NoteHeaderActions` (Ask June handler, working dot, copy token), and Rust unit tests for the mic-duck (aligned-silence-not-dropped, pause precedence) and the dictation→duck event mapping.

- [x] Tested visually where it matters — iterated live in the browser against screenshots throughout. Native app QA still recommended (see below); no screenshot attached here.
- [ ] Needs a June API (backend) deploy before it works end to end. **No** — frontend + Tauri only; the chat rides existing `/v1/*` via the embedded Hermes runtime.

## Out of scope

- No slide-out **exit** animation on either panel — the note chat panel matches the files panel (both unmount instantly while the card slides back). Adding a real exit would be a paired change to both.
- Category chips in the note composer aren't wrapped as issue reports (harmless; the panel is note Q&A).
- Mid-conversation model switch applies to the next message only, never mid-turn.

## Followups

- Native, on-device QA of the audio paths (mic-duck across pause/resume and stop-mid-dictation) and the toolbar overflow popover inside the sticky bar — the browser can't exercise the Tauri/CoreAudio runtime.
- Optional: adversarial waveform profiling to confirm the coalescing fully resolves the under-load jitter (root trigger is CPU contention, which isn't reproducible off-device).
- Optional: paired slide-out exit animation for the files + note chat panels.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/646"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1785999254&installation_id=144203540&pr_number=646&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F646&signature=1de64ad30eadb5f97e496e14f71da03a9b88a1b1f3efb53090b429f5caa4f05a"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


andrewwashuta:

**Addressed (Greptile)**
- Concurrent sends duplicate prompts → `submit()` now takes a synchronous in-flight ref guard, released in `finally`, so a double send can't create the session or reach `prompt.submit` twice before `setWorking(true)` commits.
- Session registration can vanish → consolidated bridge registration into a single best-effort `ensureHermesBridgeSession` point that retries a swallowed earlier attempt on later sends (tracked by a `bridgeEnsuredRef`), so "Open in agent view" always resolves the session.
- Terminal events can be dropped → the gateway event handler now clears `working` on an untagged terminal (error/lifecycle) frame. This gateway only ever serves the one active note chat, so a session-id-less terminal event can only mean the in-flight turn ended; untagged non-terminal events are still dropped.

**Addressed (Octopus)**
- Test byte assertion self-documented via a `BYTES_PER_SAMPLE` constant tied to the 16-bit PCM writer.

**Not changed**
- Octopus's "`noteHasBreadcrumb` removal leaves `data-has-detail-bar` always-true" is intentional — every note now renders the toolbar so its content starts at a consistent height (that's the alignment fix).

Please take another look.

andrewwashuta:
@greptileai Pushed a fix for the P1.

**Addressed**
- Registration still vanishes (open-in-agent path) → `handleOpenNoteChatInAgent` now awaits `ensureHermesBridgeSession` (best-effort, 2.5s timeout) before switching to the agent view, so the handoff guarantees the session is in the bridge list even if the first-message registration failed and no second note-chat message was sent. The submit-time retry stays as the self-heal for the panel's own continued use.

Please take another look.

andrewwashuta:
@greptileai Pushed fixes for both handoff findings.

**Addressed**
- Handoff still opens missing session (App.tsx) + First handoff races session creation (NoteChatPanel.tsx) → both resolved with register-before-expose: `useNoteChat` now exposes a reactive `bridgeRegistered` flag, and the panel **disables "Open in agent view" until the session is confirmed registered in the bridge list** (`escalationPending = (working || storedSessionId) && !bridgeRegistered`). So it can never hand the agent view an `undefined` id (creation race) or an unregistered id (transient ensure failure). App's handoff drops its own best-effort ensure and switches straight in, since the button now guarantees a registered session.

Also folded in unrelated UI feedback: the @ note suggestion rows are now uniform single-line (lead with the note's identifying text, distinct preview inline to the right, date on a right rail) — fixes the mixed one-/two-line look.

Please take another look.

andrewwashuta:
@greptileai Pushed a fix for the registration dead-end — and corrected the underlying model.

**Addressed — Registration Failure Dead-Ends**
The agent view resolves a conversation by **session id** (`listSessionMessagesOrdered(id)`), and `ensureHermesBridgeSession` is best-effort (the agent workspace's own submit path swallows its failure too). So registration only affects the **history sidebar**, not whether the conversation opens. Gating the handoff on registration was the wrong model — hence the dead-end.

- Dropped the `bridgeRegistered` gating. The panel now disables "Open in agent view" only during the pure first-send race (session being created, no id to hand off yet) — which also still covers the earlier "passes `undefined`" race.
- The handoff switches straight in by id and **backfills** bridge registration in the background (best-effort, non-blocking), so the session still shows in the agent history sidebar even if the note chat's own registration hadn't landed. It never stalls or dead-ends, and it can't open an empty conversation because the messages load by id.

This is the last planned change on my side; the author will take the final review and merge from here. Thanks for the thorough passes.

PR #659, opened 2026-07-08
fix: stream inference responses through June API to stop 504s on long prompts (JUN-225)

jakubswierczek:
Closes JUN-225

## What changed

- **June API — agent chat streams through** (`/v1/chat/completions` with `stream: true`): upstream SSE bytes are forwarded to the client as they arrive instead of being buffered until inference completes. While the upstream is silent (long prompt evaluation), June API emits SSE comment heartbeats (`: keep-alive`) so no proxy hop sees an idle response. The charge settles after the stream ends, from the usage frame in the final SSE event (`stream_options.include_usage` was already injected; now it is actually consumed from a stream). Buffered (`stream` absent/false) behavior is unchanged.
- **June API — opt-in SSE keep-alive for note generation** (`POST /v1/notes/generate` with `"stream": true`): responds `200 text/event-stream` immediately, heartbeats while generation runs, then a single terminal event carrying the exact buffered JSON (`event: result`) or the exact buffered error envelope plus its status (`event: error`).
- **Desktop — note generation client** requests the SSE mode and branches on the response content type, so it works against both deployed and not-yet-deployed backends.
- **ADR 0013** records the decision (stream instead of raising proxy timeouts; charge-after-stream; billing semantics for interrupted streams).

## Root cause

June API sits behind an nginx-based ingress (dstack-ingress) inside the TEE. Every inference response was fully buffered server-side — even `stream: true` agent chat read the whole Venice body via `response.bytes().await` before answering — so June API emitted zero bytes while inference ran. Long prompts exceeded the ingress read timeout, nginx answered `504 Gateway Time-out`, and the embedded Hermes runtime retried three times into the same wall, producing the reported "API call failed after 3 retries: HTTP 504". Raising proxy timeouts was rejected: the outer Phala gateway's limits are not ours to configure, and any fixed ceiling just moves the cliff.

## Billing semantics (deliberate decisions)

- Client disconnects mid-chat-stream → the provider keeps draining the upstream to capture the usage frame and settles the charge normally (matches buffered semantics, where a disconnect after send still charges). Generate disconnects cancel uncharged (all-or-nothing delivery, buffered parity).
- Stream completes cleanly without a usage frame → settle at the flat authorize estimate clamped to the hold cap, with a loud error log (content was delivered; June does not silently absorb upstream cost). Stream dies on a transport failure → settle nothing (the buffered path errors before its charge line on the same failure).
- Hold TTLs now cover the settle-after-upstream window: agent chat and note generation run on the same bounded metered-inference client as images (route minus authorize + settlement budgets = 390s) and their shared hold went 300s → 540s with a config-load validation pinning the relationship. Without this, any 300-600s call — exactly what streaming un-masks — would settle against an expired hold (June pays Venice, user sees `metering_provider_failed`). No deploy env overrides the TTL; the new value ships in config.toml with the image.
- The chunk channel is unbounded so settlement can never be stalled by a slow reader; worst-case memory equals the buffered path's old profile.

## Needs June API deploy

Yes. The agent-chat fix is entirely server-side. The desktop generate change is safe to ship first: old backends ignore the unknown `stream` field and answer buffered JSON, which the client still handles.

## Validation

- `make verify` green on the full branch (Biome, typecheck, vitest, fmt/clippy/test for both Rust crates).
- Review battery (high-stakes sizing: full battery, per-chunk adversarial rounds, adversarial on both harnesses):
  - Per-chunk adversarial (Codex) on the chat chunk found a real bounded-channel settlement stall (fixed: unbounded channel) and the settlement-durability gap (pre-existing parity, recorded in ADR 0013).
  - Per-chunk adversarial (Codex) on the generate chunk found a blocked-send path that suspended timeout enforcement (fixed), the disconnect-cancellation semantics question (resolved: parity with buffered), and Retry-After loss in streamed errors (fixed server-side).
  - Final battery: Standards flagged two CONTEXT.md `_Avoid_` violations in the ADR wording (fixed); Spec flagged wall-clock vs silence-gated heartbeats (fixed: interval resets on chunk arrival); Codex adversarial flagged the desktop dropping `retry_after_secs` (fixed) and idempotency-key reuse across retried attempts (pre-existing parity — key format is byte-identical on main; follow-up below); Claude adversarial found the two biggest defects of the whole cycle — the 300s-hold-vs-600s-window expired-hold billing leak and the transport-failure-vs-buffered charge divergence (both fixed, see billing semantics above). A final Codex convergence round ran over the fix commit.
- Live evidence (local june-api in LOCAL_DEV mode against a stub upstream that stays silent 22s, then trickles tokens):
  - Streamed chat: heartbeats at t+10.0s and t+20.0s, first token forwarded at t+22.0s, tokens incremental, usage frame last, charge settled from the streamed usage (settle log line confirmed). The old code would have emitted its first byte at t+29s.
  - Streamed generate: `: keep-alive` at t+10.0s, `event: result` at t+15.0s carrying the exact buffered JSON.
  - Buffered regressions: generate without the flag and chat with `stream:false` return plain `application/json`, byte-identical envelopes.
- No app-level video walkthrough: the fix is server-side + a fallback-safe client change; the failing condition (a production ingress timeout) cannot be reproduced against a local backend. The curl timelines above are the demonstrative evidence.

## Out of scope / followups

- `/v1/transcribe`, `/v1/dictate`, and image generation remain buffered; they are bounded today (chunked audio, image time budgets). If one starts hitting proxy read timeouts, extend it with the same opt-in SSE envelope (ADR 0013 consequences).
- No durable pending-charge ledger: a June API process death mid-stream loses that charge (hold expiry protects the user). Same statelessness the buffered path has; recorded in ADR 0013.
- Internal `stream: false` agent-chat callers (title suggestion, image prompt classification, permission explainer) are short prompts and were deliberately not force-streamed.
- Agent-chat charge idempotency key (`agent_chat:{user}:{model}:{body_digest}`) treats a retried identical body as a replay, so the second upstream run settles against the first charge. Pre-existing behavior (key format unchanged from main), slightly more visible with streaming; a per-attempt operation id (like the image path's request ledger) is the fix if we want it.
- No retry affordance in the chat UI for a failed turn; users resend manually. Candidate follow-up issue.

## Assumptions flagged for review

- Hermes's OpenAI-compatible client tolerates SSE comment lines (spec-legal; comments are the standard keep-alive mechanism).
- serde-based old clients tolerate the new optional `stream` request field being ignored (verified: no `deny_unknown_fields` in the api crate).

https://claude.ai/code/session_01GCtJnLvvREbhdbq6QgAYHE

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/659"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786096303&installation_id=144203540&pr_number=659&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F659&signature=46e49e40c541d10e1f7e097353cd81881d6a7d87bfc9674f14bc9be65d32c3da"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


jakubswierczek:
Re Greptile's **Stream Timeout Truncates Output** (P2): correct, and fixed in 42494ac0.

The 390s cut itself is forced — settlement must land inside the 540s authorization hold, which is capped at 600s by OS Accounts, so a stream cannot both run unbounded and settle inside a hold. What was wrong is the billing classification: the pump treated our own client timeout like an upstream fault and settled nothing after delivering up to 390s of content. Now `error.is_timeout() && !accumulated.is_empty()` classifies the deliberate cut as delivered-but-unmetered and settles at the flat estimate (same rule as a missing usage frame); a cut before any byte, or a genuine upstream failure, still charges nothing (buffered parity). Regression test: `agent_chat_stream_client_timeout_after_delivery_resolves_completed_without_usage`. ADR 0013 updated.

jakubswierczek:
@codex review

jakubswierczek:
Both Codex findings are correct and fixed in `1f78631a` (commit is local pending an SSH agent restart, will be pushed shortly):

1. **Cancel dropped generate streams before charging** — the select loop now watches `tx.closed()`, so a client disconnect cancels generation immediately instead of at the next keep-alive tick. The undelivered-result-still-charges window is gone; all-or-nothing semantics now hold exactly.
2. **Keep BYOK text calls on the full-timeout client** — `AgentChatRequest`/`GenerationRequest` now carry `unmetered` (set by the services from the same predicate that skips OS Accounts metering), and the Venice provider selects the full-route client for those calls. The shortened metered window exists only to keep settlement inside the authorization hold; BYOK calls have no hold and keep the pre-existing 600s budget.

jakubswierczek:
@codex review

jakubswierczek:
@greptile-apps please review the latest commits (42494ac0, 1f78631a) — they address your Stream Timeout finding and Codex's two P2s.

jakubswierczek:
Re Codex's **Only inject heartbeats between SSE events** (P1): correct — reqwest chunk boundaries are arbitrary bytes, and a stall after a partial `data:` line would have spliced the comment mid-frame. Fixed in `4f347067`: the pump tracks whether the last forwarded byte ended a line and the heartbeat branch is gated on that boundary (an SSE comment is legal at any line start, including between the data lines of one event). A stall after a partial line now suppresses heartbeats until the line completes — a rare, short window that risks a proxy timeout, which beats corrupting the frame. Regression test: `agent_chat_stream_heartbeat_never_splices_into_a_partial_line` (splits a data line across chunks with a stall past the heartbeat interval, asserts the line reaches the client intact).

@codex review

jakubswierczek:
Re Codex's **Avoid blank-line heartbeats inside open SSE events** (P2): correct — the `\n\n` heartbeat carried a blank line, and a blank line dispatches whatever event fields are already buffered, so a heartbeat between the data lines of one legal multi-line event would split it. Fixed in `2c5e09d1`: heartbeats are now a lone comment line (`: keep-alive\n`) in both the chat pump and the generate handler — comments never dispatch, and proxies only need bytes. Combined with the line-boundary gate from `4f347067`, a heartbeat can now appear only where the SSE grammar makes it invisible.

@codex review

jakubswierczek:
Re Codex's **Cancel BYOK streams when the downstream drops** (P2): correct — the drain exists to capture the usage frame for June's settlement, and BYOK has no settlement, so draining only spent the user's own upstream account on a response nobody reads. Fixed in `675b4601`: the pump carries `drain_for_settlement` (false for unmetered calls) and aborts on the first failed send after a disconnect, dropping the upstream connection. Detection is bounded by the heartbeat tick during silence. Regression test: `agent_chat_stream_unmetered_aborts_instead_of_draining_after_disconnect`.

@codex review

jakubswierczek:
Re Codex's **Prioritize disconnects before completing streamed generation** (P2): correct — the select was unbiased, so a disconnect and a ready generation in the same wake could resolve in favor of the result, charging for an undeliverable response. Fixed in `d891931b`: the select is `biased` with `tx.closed()` polled first, so a visible disconnect always wins and the generation future is dropped before it can be driven across its charge call. One honest caveat, noted in the code: bias narrows the race to the unavoidable in-flight window — a charge request already on the wire cannot be recalled client-side.

jakubswierczek:
Re Codex's **Watch BYOK downstream closure while upstream stalls** (P2): correct — the two heartbeat-safety fixes (line-boundary gate, lone comment line) had quietly removed BYOK's only disconnect-detection channels during a mid-line stall, and non-SSE bodies never had one. Fixed in `36f07035`: the pump adds a `chunks_tx.closed()` branch gated on `!drain_for_settlement`, so BYOK aborts the upstream the moment the downstream drops, independent of data flow. Metered streams deliberately keep draining (settlement needs the usage frame). Regression test: disconnect during a partial-line stall resolves `Failed` immediately instead of waiting out the stall.

jakubswierczek:
Promoting the final-battery **idempotency key reuse across retried identical bodies** finding from follow-up to fixed here, in `36b96880`: both agent chat paths now settle under an attempt-scoped key (`agent_chat:{user}:attempt:{operation_id}:{body_digest}`, the image scheme, operation id minted before the upstream call). A retried identical body runs a fresh upstream completion that delivers its own content, so it settles its own charge — the content-derived key let that second settlement replay as a no-op. Deliberately NO retry-dedup ledger in front (unlike images): chat has no client-supplied request id, and any retry that reaches the upstream does real billable work. `note_generate` keeps its content-derived key — its dedup is load-bearing for all-or-nothing billing on ambiguous retries. ADR 0013 updated in the same commit.

@codex review

jakubswierczek:
@codex review

jakubswierczek:
@codex review

Merged latest `main` (through #666) into the branch and resolved the sole conflict in `june-api/crates/api/tests/http_boundary.rs` — unified the streaming-test `TestStateDeps` builder with main's P3A `p3a_sink` injection, keeping both the streaming and P3A integration tests. Full local gate green except one pre-existing `onboarding.test.tsx` failure that lives on main's HEAD (telemetry-consent heading wording vs assertion) and is unrelated to this PR.

jakubswierczek:
@greptile-apps please re-review the latest commit (`38dc278b`): merge of `main` into this branch plus the `http_boundary.rs` conflict resolution. Your prior 5/5 was on `675b4601`, several commits back — this brings you current through the two Codex P2 fixes, the idempotency-key fix, and the main merge.

jakubswierczek:
@codex review

jakubswierczek:
@greptile-apps please review the current tip `f58d9105` — identical tree to the prior `38dc278b` (only the commit signature changed).

PR #701, opened 2026-07-10
Attach multiple issue report videos without silent drops (JUN-238)

junhohong:
Closes JUN-238

## Root cause

The multi-file UI was already passing both names and paths. The loss happened later in the June-owned path:

- Tauri silently omitted any attachment over its old 10 MiB per-file limit.
- June API inherited the unrelated 25 MiB audio request limit.
- Production ingress was capped at 80 MiB.
- Open Software attachment failures were only logged, so neither the user nor the resulting Issue named the missing file.

This belongs in `os-june`, not a new `os-platform` PR. os-platform PR #100 already raised generic files to 300 MiB, and ADR 0012 requires desktop reports to flow through authenticated June API before June API files the Open Software Issue.

## What changed

- Keep native Add files selections path-based instead of copying them through the generic 50 MiB Hermes importer, then stream every readable, non-empty attachment that fits the 300 MiB per-file and cumulative budget. Preserve MP4, MOV, M4V, and WebM MIME types.
- Keep WKWebView DOM drops bounded to 50 MiB before their unavoidable byte import, process all accepted files sequentially up to the API's 20-file contract, and show visible guidance for size/count violations because DOM `File` objects expose no filesystem path.
- Keep skipped filenames in the report and confirmation for unreadable, empty, over-budget, compatibility-fallback, and downstream-rejected files.
- Opt into a heartbeat SSE response while June API transfers attachments and files the Issue, preventing the nginx/Phala idle-response timeout. Old clients keep the buffered JSON contract, and new clients accept either format.
- Retry once with names only when an older June API or ingress rejects the multipart size, so a staged desktop/API rollout does not lose the report.
- Give June API a dedicated 301 MiB issue-report body budget and production ingress a 302 MiB limit.
- Limit the complete large-report lifetime, from body extraction through downstream delivery, to one concurrent request. Retain attachment bytes in shared `Bytes` and stream the same allocation to os-platform without a second platform-sized clone.
- Use one 600-second server deadline from permit wait through terminal SSE, with a dedicated 900-second desktop window for connection, request-transfer, and terminal-response grace. Explicitly disable protocol retries and avoid the project-to-org fallback after ambiguous transport or malformed-envelope failures; org fallback remains only for a parsed project rejection.
- Return downstream unattached filenames to the desktop and add them to the resulting Open Software Issue body.

## Validation

- `make verify`: green.
  - Frontend: 144 test files, 2,318 passed, 2 skipped.
  - Tauri: rustfmt, Clippy with warnings denied, and all tests passed.
  - June API: rustfmt, Clippy with warnings denied, and all workspace tests passed.
- Focused coverage crosses the former 10 MiB boundary, checks the 300 MiB per-file/cumulative budgets with sparse files, sends repeated multipart attachment parts, verifies two Open Software file transfers and file IDs, exercises structured 413 handling, checks SSE/JSON compatibility and heartbeat behavior, holds the concurrency permit through disconnected streamed delivery, prevents ambiguous Issue replay, and propagates downstream failures.
- Live QA used the real `ReportDialog` with a stubbed Tauri bridge: two video attachments remained visible, both names and paths were submitted, and the dialog reached its success state. The latest QA artifact is linked in the PR comments.

## Deployment and scope

Deploy June API before relying on the higher end-to-end limit. Newer desktop builds remain compatible with older API deployments through the names-only fallback; older desktop builds ignore the additive response field.

No os-platform change is needed: its 300 MiB file capacity and video playback support are already merged. A future streaming/pre-attachment API could reduce June API's single-request memory floor further, but that is a separate wire-contract change.


junhohong:
QA video artifact

- Video: https://app.opensoftware.co/api/v1/files/fil_5jcO6XEwakaU/download
- Compressed file: `jun-238-report-dialog-qa.compressed.mp4` (317.5 KiB)

junhohong:
@codex review

junhohong:
QA video artifact

- Video: https://app.opensoftware.co/api/v1/files/fil_gB55FWldB0TI/download
- Compressed file: `jun238-two-video-submit.mp4` (75.3 KiB)

junhohong:
Implementation update: the earlier QA walkthrough that blocked files over 10 MB is obsolete. The final branch keeps native picker paths directly, streams files within the 300 MiB report budget, and uses the June API heartbeat response through Open Software delivery.

Latest synthetic UI QA: https://app.opensoftware.co/api/v1/files/fil_gB55FWldB0TI/download

@codex review

junhohong:
Octopus finding disposition: no code change. `/v1/issue-reports` currently produces only buffered JSON bytes or a plain SSE byte stream, and no middleware on this route adds trailers. The `into_data_stream()` wrapper is intentional so the shared large-report permit stays alive through the body stream. If this route later gains trailer-producing middleware, it should move to a frame-preserving body wrapper as part of that change.

PR #702, opened 2026-07-10
Add queued steering controls to the agent composer

andrewwashuta:
## Summary

- Show working-composer follow-ups in one unified, collapsible "Up next" surface with a combined count.
- Reuse the compact horizontal steering-row treatment for text and attachment follow-ups; row status distinguishes "Steering current turn" from "Waiting for June to finish."
- Keep submitted text steers read-only because Hermes cannot recall them, while locally waiting attachment messages retain honest Edit and Remove actions.
- Send queued attachment follow-ups one turn at a time after clean completion, behind any undrained text-steer fallback, and retain failed deliveries for retry.
- Give image and file attachments the same 44px treatment with a subtle file surface and themed document icon.
- Add a dev-only `__upNextDemo()` steering preview that renders the mixed state without submitting a message.

## Validation

- `pnpm typecheck`
- `pnpm check` (passes with existing repository warnings only)
- Full frontend suite: 143 files passed; 2,323 tests passed and 2 skipped.
- Added regressions for the unified surface, no-submit preview, local edit/remove, delayed file and image delivery, FIFO serialization, duplicate completion, continuity identities, retry, and partially attached images.
- Browser-verified the combined horizontal state, attachment sizing, action alignment, and card density.
- Commit-specific `signoff/frontend` and `signoff/rust-macos` checks passed on `fdbe4ae8`.

- [x] Does not need a June API (backend) deploy to work end to end.

## Root cause

The working composer exposed immediate text steering and delayed attachment delivery as two competing queue surfaces. That duplicated the same user intent, obscured which controls were honest, and produced an oversized attachment card even though both paths are follow-ups to the active session.

## Out of scope

- Recalling, mutating, or reordering a text instruction after Hermes accepts it.
- Persistent submitted-steer history after the turn finishes.
- Steering support in the note-side chat panel.
- An attachment lightbox.

## Followups

- Consider a separate attachment-lightbox story if full-size previews and filenames are needed.

CleanShot 2026-07-10 at 17 05 45@2x

CleanShot 2026-07-10 at 17 06 34@2x


andrewwashuta:
@greptileai I pushed changes addressing the review.

**Addressed**
- Removed edit/remove semantics from submitted steers. Rows are now read-only because `session.steer` cannot recall or mutate an instruction after submission.
- Preserved pending delivery tracking independently from the visible rows and added a regression covering terminal fallback delivery.
- Kept the immediate Stop visual but disabled it until the runtime session is actually marked working, preventing an early interrupt race.
- Merged current `main`, preserving both the new video-generation placeholder precedence and steering-specific composer state.

**Skipped (with reason)**
- Preserving an edited steer in its original position: submitted rows are now immutable. Any correction is a new steer and intentionally appends in delivery order.

Validation: typecheck and Biome pass; full frontend suite 2,311 passed / 2 skipped; focused suite 258 passed / 2 skipped.

Please take another look.

andrewwashuta:
@greptileai Final confirmation requested: the post-review change only scopes composer submit state to its owning Hermes session and adds deterministic regression coverage. Local signoff is complete. Please confirm there is no regression.

andrewwashuta:
@greptileai The remaining finding is fixed at HEAD.

**Addressed**
- "Scope Stop state" (composer render path bypassing the session check) → the inline `workingSessionIds.has(...) || submitting || composerSteerDemo` condition was replaced by the shared `composerInSteerStateFor` helper, which only counts an in-flight submit when `submittingSessionId === selectedSessionId`. Covered by tests asserting a submit owned by another session does not put the selected idle session into the Stop/steer branch.

Since your last review the branch also picked up a series of visual-polish commits on the "Up next" queue (shared leading rail, concentric insets, compact header triggers, white file tiles, multi-attachment files glyph) plus an overflow fix that stops the queue list from panning horizontally. No API or behavior contract changes.

Please take another look.

andrewwashuta:
@greptileai Pushed one more commit addressing the two substantive Codex findings; please confirm it doesn't change your verdict.

**Addressed (Codex)**
- Queued follow-up preparation failures → the attachment-queue branch now catches `prepareComposerSubmission` rejections and shows the recoverable error banner (draft and attachments stay in the composer).
- Failed rows auto-resent by queue advancement → automatic delivery after a clean turn now stops at a failed head; only the row's explicit Retry/Send resends it.

**Skipped (with reason)**
- "Release the continuation guard before starting follow-ups" (Codex) → theoretical race requiring a turn to complete faster than the submit's session reload; the guard releases in `finally` immediately after, and restructuring risks double-delivery, which is worse than the rare self-healing stall.
- "Do not mark possibly accepted follow-ups as failed" (Codex) → the restore-after-unmount state is genuinely ambiguous (accepted vs not); fixing it properly needs an accepted-at-ack state threaded through `submitHermesSession`, a larger change we're deferring rather than guessing.

Regression tests added for both fixes; full suite green (261 passed).

PR #728, opened 2026-07-13
Auto toggle, preference drill-in, and preset readback on the composer pill

andrewwashuta:
## Summary

- **Auto toggle in the model picker** (workspace composer and Settings → Models): a pinned section above the suggested rows with a switch that selects `open-software/auto` while keeping the popover open; toggling off lands on the leading suggested pick (GLM 5.2). Suggested rows return to concrete models (GLM 5.2 / Kimi K2.6 / GLM 5.1) — this deliberately replaces the three Auto preset rows from #726 with the toggle design (product decision by @andrewwashuta).
- **Preference drill-in**: while Auto is on, a "Preference · Higher quality ›" row opens a side flyout with the three cost-to-quality presets as rows, each with a one-line explanation and a checkmark on the active one. Writes persist via `set_cost_quality` and announce the settings-changed event so every surface stays in sync.
- **Composer pill readback** (workspace and note chat): the pill ghosts the active preset beside the name — "Auto Higher / Balanced / Lower" — one visual step quieter than the model name in every state. Pill resting contrast steps up from muted grey toward foreground.
- **Picker layer polish**: second-layer surfaces drop `background-clip: padding-box` so their translucent border composites over the surface's own fill like the root popover (same token, no more darker second-layer border), and the preference flyout gets the catalog panel's `--sp-1` content inset.
- Note-chat designation staleness fixed per review: the panel subscribes to `june:provider-model-settings-changed` and re-reads the preference; its own preset writes dispatch the same event for the mirror direction.

## Root cause

(Border fix) `.agent-composer-model-surface` set `background-clip: padding-box`, so its translucent `--popover-border` blended over the backdrop behind the panel (page + neighboring panel's shadow halo) rather than over the panel's own fill like the root popover — same token, darker composite.

## Validation

- `pnpm typecheck`, Biome clean on changed files; 354 tests pass across `agent-workspace`, `app-settings`, `suggested-models`, `model-picker-dialog`, `note-chat-sessions`, and `hermes-model-switch`, including rewritten coverage: toggling Auto on from the picker, drilling into the preference flyout, and the pill designation tracking the persisted value.

- [x] Tested visually where it matters (toggle, drill-in, pill, and border treatments were iterated live in the dev app during design review).

## Out of scope

- The note chat panel's own picker popover (duplicated component) doesn't render the Auto toggle section yet — its pill shows the designation and stays in sync, but preference editing happens from the workspace picker or Settings. Consolidating the duplicated popovers is a natural follow-up.
- Shifting the Auto preference for in-flight work — the preference is read per request today, so streams in progress keep their routing decision.

## Followups

- JUN-306: backend contract for changing the Auto routing preference while work is in flight.
- Consolidate `ComposerModelPopover` (note chat) with the shared `ModelPickerPopover` so the Auto section exists in one place.

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -248,8 +248,12 @@ import {
   rawLocalGenerationModelId,
   withLocalGenerationOption,
 } from "../../lib/local-generation";
-import { preferredVisionFallbackModel } from "../../lib/suggested-models";
-import { modelOptions, selectedModel as selectedModelOption } from "../settings/ModelPickerDialog";
+import { autoPillDesignation, preferredVisionFallbackModel } from "../../lib/suggested-models";
+import {
+  AUTO_MODEL_ID,
+  modelOptions,
+  selectedModel as selectedModelOption,
+} from "../settings/ModelPickerDialog";
 import { ModelPickerPopover, type ModelPickerFlyout } from "../settings/ModelPickerPopover";
 import {
   HERMES_SERVER_ERROR_MESSAGE,
@@ -9337,6 +9341,11 @@ export function AgentWorkspace({
               <ComposerModelPicker
                 open={composerModelOpen}
                 model={generationModel}
+                detail={
+                  generationModel?.id === AUTO_MODEL_ID
+                    ? autoPillDesignation(generationCostQuality)
+                    : undefined
+                }
                 readOnly={composerModelLocked}
                 triggerRef={composerModelTriggerRef}
                 onToggleOpen={() => {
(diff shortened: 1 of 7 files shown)


andrewwashuta:

**Addressed**
- Note-chat Auto designation becomes stale (P1) → the panel now subscribes to the `june:provider-model-settings-changed` event and re-reads the persisted preference when a generation change is announced elsewhere. Also fixed the mirror case: the panel's own preset writes now dispatch the same event, so the workspace pill refreshes too.

Please take another look.

andrewwashuta:
@greptileai I've pushed a substantial follow-up: the Auto preset suggested rows are replaced by the Auto toggle + preference drill-in design (product decision — see the updated PR description). Suggested rows return to concrete models, Auto lives in a pinned toggle section with a drill-in flyout for the cost-to-quality preference, and the tests were rewritten for the new flow.

Please take another look at the full diff.

PR #732, opened 2026-07-13
Restore the Auto toggle and preference drill-in to the model picker

andrewwashuta:
## Summary

- Replaces the three "Auto · …" preset suggested rows (#726, #730 copy) with the **Auto toggle design** (product decision by @andrewwashuta): suggested rows return to concrete models (GLM 5.2 / Kimi K2.6 / GLM 5.1), and Auto lives in a pinned section above them.
- **The toggle** selects `open-software/auto` while keeping the popover open; toggling off lands on the leading suggested pick (GLM 5.2). Present in the workspace composer picker and Settings → Models. The catalog's Auto entry is filtered out of the suggested/All-models lists so the toggle is the router's one home.
- **The Preference drill-in**: while Auto is on, a "Preference · Higher quality ›" row opens a side flyout with the three cost-to-quality presets as rows (one-line explanation each, checkmark on the active one). Writes persist via `set_cost_quality` and dispatch the settings-changed event so the note-chat pill designation stays in sync (mirror of the sync shipped in #728).
- Composes with what already merged: the pill designation readback and picker border fixes (#728) and the Auto branding/spec-hiding (#730) are untouched and keep working with the toggle.

## Validation

- `pnpm typecheck`, Biome clean; 359 tests pass across seven suites, including #730's Auto-branding tests and rewritten coverage: toggling Auto on from the picker (popover stays open), drilling into the preference flyout, the pill tracking the persisted preset, and `suggestedModelsForMode` never surfacing Auto as a row.

- [x] Tested visually where it matters (toggle, drill-in, and flyout inset were iterated live in the dev app during design review).

## Out of scope

- The note chat panel's own picker popover (duplicated component) doesn't render the Auto toggle section yet — its pill shows the designation and stays in sync, but preference editing happens from the workspace picker or Settings.
- Shifting the Auto preference for in-flight work (preference is read per request today).

## Followups

- JUN-306: backend contract for changing the Auto routing preference while work is in flight.
- Consolidate `ComposerModelPopover` (note chat) with the shared `ModelPickerPopover` so the Auto section exists in one place.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/732"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1786571253&installation_id=144203540&pr_number=732&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F732&signature=833a251b7c50b9426bb30881ced66829bff23a85b7e3b02cc066de30b66157dc"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/components/agent/AgentWorkspace.tsx
+++ b/src/components/agent/AgentWorkspace.tsx
@@ -2540,9 +2540,12 @@ export function AgentWorkspace({
   const [generationCostQuality, setGenerationCostQuality] = useState<number | undefined>();
   // Preference saves from the picker's drill-in: writes are chained so they
   // persist in click order, and versioned so only the newest call's outcome
-  // touches the UI (mirrors Settings' saveCostQuality discipline).
+  // touches the UI (mirrors Settings' saveCostQuality discipline). Rollback
+  // targets the last CONFIRMED value (persisted read or successful save) —
+  // never an optimistic value a still-in-flight click painted.
   const costQualitySaveChainRef = useRef<Promise<unknown>>(Promise.resolve());
   const latestCostQualitySaveRef = useRef(0);
+  const confirmedCostQualityRef = useRef<number | undefined>(undefined);
   const defaultGenerationModelIdRef = useRef("");
   const [generationModels, setGenerationModels] = useState<VeniceModelDto[]>([]);
   const generationModelsRef = useRef<VeniceModelDto[]>([]);
@@ -3593,6 +3596,7 @@ export function AgentWorkspace({
       setGenerationProvider(provider);
       defaultGenerationModelIdRef.current = selectedModelId;
       setDefaultGenerationModelId(selectedModelId);
+      confirmedCostQualityRef.current = settings.costQuality;
       setGenerationCostQuality(settings.costQuality);
       return selectedModelId;
     },
@@ -3751,7 +3755,6 @@ export function AgentWorkspace({
     // newest call's outcome (success or rollback) touches the UI — the same
     // discipline as Settings' saveCostQuality.
     const version = ++latestCostQualitySaveRef.current;
-    const confirmed = generationCostQuality;
     setGenerationCostQuality(value);
     const save = costQualitySaveChainRef.current.then(() => setCostQuality(value));
     costQualitySaveChainRef.current = save.then(
@@ -3760,6 +3763,7 @@ export function AgentWorkspace({
     );
     void save.then(
       (next) => {
+        confirmedCostQualityRef.current = next.costQuality;
         if (version !== latestCostQualitySaveRef.current) return;
         setGenerationCostQuality(next.costQuality);
         dispatchProviderModelSettingsChanged({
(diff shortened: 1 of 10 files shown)


andrewwashuta:

**Addressed**
- Stale preference save wins (P1) → the workspace drill-in now chains writes (persist in click order) and versions them (only the newest call's success or rollback touches the UI), mirroring Settings' saveCostQuality discipline.
- Empty catalog traps Auto on (P1) → toggle-off falls back to the exported curated default id (`DEFAULT_GENERATION_SUGGESTION_ID`, GLM 5.2) when no concrete model is selectable; selecting an id absent from the catalog is safe (the pill renders a name-only stub).

Please take another look.

andrewwashuta:
@greptileai On the remaining finding (empty-catalog fallback persists a model absent from the live catalog) — I'm keeping the current behavior deliberately, and here's the reasoning:

- `DEFAULT_GENERATION_SUGGESTION_ID` mirrors `DEFAULT_GENERATION_MODEL` in the Rust providers module — it's the id the backend itself defaults to and sanitizes toward for fresh installs. Persisting it returns settings to the factory default, not to an arbitrary unavailable model.
- An empty catalog at toggle time is a transient fetch state (bridge starting, network hiccup), not evidence the model is unusable upstream: generation goes through June API regardless of whether the catalog request succeeded. Validating the selection against an absent catalog is impossible by definition — any choice here is sight-unseen, including staying on Auto.
- The alternative failure mode is the one flagged in the previous round: a checked switch that cannot be turned off. Between "switch stuck on Auto" and "settings return to the app default, pill shows a name-only stub until the catalog loads", the second is strictly less surprising, and it self-heals as soon as the catalog arrives.

If the default id were ever delisted, the existing stale-model routing on main (open-software/auto fallback for stale text models) catches it at request time.

No code change in this round. Please take another look with that context.

andrewwashuta:
@greptileai I've pushed a change addressing the remaining finding.

**Addressed**
- Turning Auto off with an empty catalog persists an unavailable model → the blind default-id fallback is gone. Toggle-off now only lands on a model the catalog vouches for (leading suggested pick, else first selectable), and while none exists the switch renders disabled with an explanatory accessible name — no silent no-op, no sight-unseen persistence. It re-enables as soon as the catalog loads.

Please take another look.

andrewwashuta:
@greptileai I've pushed a fix for the rollback finding.

**Addressed**
- Rollback restores optimistic state → added `confirmedCostQualityRef`, updated only from persisted reads (`commitGenerationSettings`) and successful saves; the failure path now rolls back to that instead of the call-time React state. Two queued failing saves land the UI back on the last value that actually persisted.

Please take another look.

PR #743, opened 2026-07-14
Redesign the generating-media placeholder: themed particle field + develop-in reveal

andrewwashuta:
## Summary

Redesigns the generated image/video **generating** state and its completion into one coherent, on-brand experience.

- Generating state is a canvas particle **dot field** carrying the June Agents mark. The mark **twinkles in the active theme accent** (deterministic per-dot glint), a **pointer press bursts theme color** outward through the field, and the ambient sheen sweeps diagonally at the shared shimmer angle.
- The gray canvas **gradates into the card surface** at the bottom (dots thin out over the same run), so the footer transition is owned by the background instead of an overlay fading in on top.
- **Completion reveal:** the finished image/video **develops out of the field** — the dot surface mounts over the media (its entrance ripple doubling as the burst) and dissolves, then the caption chrome fades in. A shared `useGeneratedMediaReveal` hook drives both image and video.
- Completed media now reads as **one card** in the same chrome family as the generating card, with the picture **nested concentrically** (inner radius = outer − 4px inset), matching the artifact icon-tile idiom. Caption footer restyled and sized up a notch; filename lightened.
- The granted **permission pill follows the theme accent** rather than a generic green, so it sits in the earthy palette.

## Demo

Full flow — generating → press ripple → develop-in reveal → completed card:

Generating-media flow

Generating state (still):

Generating state

<sub>Media lives under `.github/pr-media/` on this branch so it embeds inline (GitHub's drag-drop attachment can't be minted from CLI). Safe to drop that commit before merge.</sub>

## Root cause

Bug JUN-305: the `MEDIA:…png` reference the model emits is a raw transport line June normally converts into the inline image and strips from prose — but that strip only ran when the message **completed**. While the reply streams, raw deltas paint verbatim, so the file link flashed as literal text before the image resolved. Fixed by stripping complete MEDIA references (and holding back a partially-streamed trailing one) at render time for still-streaming text (`stripRenderedMediaReferences` in `agent-chat-runtime.ts`); stored deltas stay verbatim because the completion pass needs them for reconciliation. A generated image also no longer gets a duplicate "Generated files" download card.

## Validation

- `pnpm check` (Biome) and `pnpm typecheck` clean; touched Vitest suites green (agent-workspace, agent-chat-runtime, agent-tool-labels, hermes-classifier, shimmer-system — 431 passing), including regressions for the streaming MEDIA strip, the placeholder, and the tool tagging.
- Exercised end-to-end in the real app (see Demo) and in a browser preview across generating → reveal → complete, in light and dark.

- [x] Tested visually — see the Demo section above.
- [ ] Needs a June API (backend) deploy before it works end to end. — **No, frontend-only.**

## Out of scope

- The file/artifact viewer and download plumbing (beyond suppressing the duplicate generated-file card).
- Video's elapsed-time progress copy — the video placeholder keeps its existing note below the frame.

## Followups

- The completion reveal is intentionally gated to live running→complete flips; history loads and reduced-motion render the finished card instantly (no follow-up needed, noted for reviewers).

- [x] Security-sensitive changes were called out in the summary. — None; frontend-only presentation change.
- [x] Workflow action references are pinned to full-length commit SHAs. — No workflow changes.
- [x] Desktop signing, updater, entitlement, capability, or release changes have owner review. — None.
- [x] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. — None.


andrewwashuta:
@greptileai I've pushed fixes for both findings.

**Addressed**
- **Global name match hides files** (`AgentWorkspace.tsx`, `assignArtifactsToTurns`) → a path-bearing inline media part now claims only its exact `path`; the basename fallback (`mediaNames`) is populated only for pathless inline media (MCP image blocks that carry a filename but no path). An inline `/generated/output.png` no longer suppresses an unrelated later `/reports/output.png`. The JUN-305 dedup is preserved — that inline part is a name-only MCP block, so it still hits the name fallback (regression test stays green).
- **Terminal scrub depends on classification** (`agent-chat-runtime.ts`, `completeRunningParts`) → the terminal MEDIA scrub now runs when a media tool ran (fast path) **or** when the text itself still carries a real `MEDIA:<path|filename>` ref, via a new `containsMediaReference` helper built on the existing anchored patterns. An unclassified media tool (e.g. `june_image`) can no longer strand a trailing `MEDIA:` line on an interrupted turn; ordinary prose containing "media" is untouched.

Checks: Biome + typecheck clean; `agent-workspace` and `agent-chat-runtime` suites green with no test changes needed. Please take another look.

andrewwashuta:
@greptileai On the follow-on ("Pathless media still collides"): I tried the per-turn scoping and it regresses the JUN-305 dedup, so I'm keeping the current fix and explaining why, rather than shipping a change that reopens the original bug.

**Why per-turn scoping doesn't work here.** A `role:"tool"` message doesn't create its own turn — `buildHermesSessionChatTurns` attaches its inline image part to the assistant turn that *made the tool call*. In the JUN-305 fixture the pathless inline image lives in the tool-call turn (`message-1`, empty text), while the backing workspace artifact is claimed by the later assistant text turn (`message-3`, the only turn with text, since the claim loop skips text-less turns). Scoping the basename suppression to the media's producing turn therefore never reaches the turn that claims the artifact, and the duplicate "Generated files" card comes back (the regression test fails). Same-turn / forward-scoping / budget-of-one all either reopen JUN-305 or fail to fully close this case, because a pathless MCP block carries *only* a basename — there's no path or directory to disambiguate "same file, different turn" from "different file, same basename."

**Residual risk is narrow.** This fallback only affects agent-generated inline images (MCP image blocks), which carry unique, effectively-hashed names (e.g. `generated-image-<hash>.png`). A real collision needs an *unrelated* workspace file whose basename exactly equals that generated name — vanishingly unlikely in practice. Greptile's original concrete example (two *path-bearing* `output.png`s) is fully fixed by this PR.

**Robust fix = a followup, not this PR.** Closing the pathless case precisely means giving the inline MCP block a resolvable identity (a path or content hash) so it dedups by exact identity instead of basename — a data-model change worth its own change with a fixture pinning the "same response run" boundary. Tracking it as a followup.

Please reconsider with the JUN-305 constraint in mind. Finding 2 (terminal MEDIA scrub) and your original path-collision example are both addressed.

PR #754, opened 2026-07-14
Reconcile generated media after gateway reconnects

andrewwashuta:
## Summary

- Reconcile stale live media-tool events with persisted Hermes tool results so reconnects leave one generation placeholder.
- Promote the pinned gateway's early id-less media placeholder to the stable tool ID when execution starts, preventing the delayed second card for image and video generation.
- Normalize completion media and correlate persisted cache paths with signed generated filenames.

## Root cause

The pinned Hermes gateway emits `message.start`, then id-less `tool.generating` while tool arguments are still streaming, and only later emits `tool.start` with the stable tool ID. Because the assistant turn already existed, the early event created a provisional media card; the later identified start used a different key and created a second card a few seconds later. Reconnects could compound this by replaying a stale live start beside a persisted result.

The runtime now preserves the already-mounted provisional card and promotes its identity when the stable start arrives. Reconnect reconciliation continues to suppress stale starts and normalize the runtime's completion shapes without collapsing genuinely concurrent generations.

## Validation

- `make local-ci` - 160 test files passed; 2,637 tests passed and 2 skipped.
- Image and video runtime regressions cover `message.start` → `tool.generating` → `tool.start`.
- AgentWorkspace renderer regression asserts the raw gateway sequence renders exactly one image placeholder.
- TypeScript, Biome, and repository signoffs passed on commit `a091ff3a`.
- Connected June API mode starts successfully with OS Accounts forwarding enabled.

- [ ] Tested visually where it matters. The issue was reproduced in the native app and screenshots; the fixed event sequence is covered end to end through the frontend classifier/runtime/renderer test, but a post-fix native screenshot was not captured.
- [ ] Needs a June API (backend) deploy before it works end to end. No June API deploy is required; this is desktop TypeScript reconciliation only.

## Out of scope

- No changes to image or video generation, pricing, model behavior, or Hermes persistence.
- No changes to OS Accounts authentication or billing contracts.

## Followups

- None.

- [x] No security-sensitive behavior changed.
- [x] No workflow action references changed.
- [x] No desktop signing, updater, entitlement, capability, or release behavior changed.
- [x] No backend auth, billing, metering, CORS, rate limit, or logging behavior changed.
- [x] No user-facing copy changed.

### DIFF
--- a/src/lib/agent-chat-runtime.ts
+++ b/src/lib/agent-chat-runtime.ts
@@ -723,6 +723,9 @@ function appendLiveHermesEvents(
           // stream once its tool reaches a terminal state.
           currentAssistant.status = "complete";
         }
+        if (hasExplicitToolIdentity && event.phase === "start" && media) {
+          promoteIdlessMediaToolPart(currentAssistant.parts, event.key, name, media);
+        }
         upsertToolPart(
           currentAssistant.parts,
           {
@@ -1158,6 +1161,29 @@ function upsertToolPart(
     status: next.status,
     ...(next.media ? { media: next.media } : {}),
   });
+}
+
+function promoteIdlessMediaToolPart(
+  parts: AgentChatPart[],
+  id: string,
+  name: string,
+  media: NonNullable<AgentChatToolPart["media"]>,
+) {
+  // tool.generating precedes tool.start: keep the early canvas mounted while
+  // replacing its provisional identity with the execution's stable tool id.
+  for (let index = parts.length - 1; index >= 0; index -= 1) {
+    const part = parts[index];
+    if (
+      part?.type === "tool" &&
+      part.id.startsWith("idless:") &&
+      part.status === "running" &&
+      part.name === name &&
+      part.media === media
+    ) {
+      part.id = id;
+      return;
+    }
+  }
 }
 
 function latestAssistantTurnWithRunningTool(
--- a/src/lib/agent-chat-runtime.ts
+++ b/src/lib/agent-chat-runtime.ts
@@ -511,6 +511,12 @@ function assistantTurnForTimestamp(turns: AgentChatTurn[], createdAt: string | u
   return undefined;
 }
 
+function liveToolCorrelationKey(event: Extract<JuneHermesEvent, { kind: "tool" }>) {
+  const name = toolActivityLabel(event.name ?? "tool", event.sanitizedPayload);
+  const media = generatedMediaToolKind(event.name, event.sanitizedPayload) ?? "none";
+  return `${name}\u0000${media}`;
+}
+
 function appendLiveHermesEvents(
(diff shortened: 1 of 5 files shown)


andrewwashuta:
@greptileai Please review this live-runtime follow-up. The regression fixtures come from the user’s real duplicate-placeholder screenshot and Hermes session database after PR #752 merged.

andrewwashuta:
@greptileai I pushed fixes for the review findings.

**Addressed**
- Removed cross-clock name/timestamp suppression. Orphan id-less media callbacks are now ignored only when there is no running matching invocation.
- Id-less callbacks correlate only to running rows, so concurrent completions resolve separate calls and terminal rows cannot be revived.
- Completion media now combines and deduplicates complementary `content`, `result`, and `structuredContent` blocks.
- Stable tool rows use the classifier event key when identity is present.

Added focused regressions for clock skew, overlapping calls, terminal-row reuse, and complementary result media. Please take another look.

andrewwashuta:
@greptileai I pushed the remaining reconnect guard.

**Addressed**
- When an explicit live tool event is suppressed because its invocation is already persisted, the replay now marks that name/media correlation as ambiguous. Later id-less callbacks for it are ignored instead of being allowed to complete a newer same-name run. Explicit events for the newer invocation still update it normally.
- Added a regression with a persisted old call, a newer running call, a delayed id-less stale completion, and the newer explicit completion; only the newer image is retained.

Focused tests, Biome, and typecheck pass. Please take another look.

andrewwashuta:
@greptileai I replaced the broad suppression set with a protocol-level guard.

**Addressed**
- The pinned Hermes gateway emits terminal media through `_on_tool_complete`, whose payload always includes `tool_id`. Its only id-less media callback is non-terminal `tool.generating`.
- June now correlates id-less media progress with a running row, but refuses to assign an ambiguous id-less terminal frame to any invocation. Explicit terminal events still complete the correct row, and hydrated history remains authoritative for older gateways.
- Updated the overlapping-call regression so ambiguous id-less terminal frames cannot complete either explicit invocation; the delayed-stale-callback regression still proves only the current explicitly identified image is retained.

Focused tests, Biome, and typecheck pass. Please take another look.

andrewwashuta:
@greptileai I found and fixed the live reproduction after the prior review.

**Addressed**
- The pinned gateway emits `message.start` → id-less `tool.generating` → identified `tool.start`. The later start now promotes the existing media placeholder to the stable tool ID instead of creating a second card.
- Added image and video regression coverage for the real gateway order.
- Updated the UI-level test to assert the raw sequence renders exactly one image placeholder.

**Verified**
- Biome targeted check
- TypeScript
- 100 agent chat runtime tests
- 296 AgentWorkspace tests (2 skipped)

Please take another look.

PR #786, opened 2026-07-16
Add seeded account avatars

andrewwashuta:
## Summary

- Replace the generic sidebar account glyph with a lightweight, theme-aware cloud Avatar without initials.
- Implement the network-wide Open Software Avatar v1 contract from merged [open-software-network/os-accounts#101](https://github.com/open-software-network/os-accounts/pull/101): supported saved seeds and derived defaults share canonical geometry across Apps, while June supplies the active theme palette.
- Add Refresh under General > Account and persist explicit selections through the User's OS Accounts `avatar_seed`.
- Keep failed writes fail-soft on the current device, guard logout and stale-response races, and explain legacy-scope failures without treating the User as signed out.
- Move Avatar action feedback out of the settings layout and into the existing top-right toast system.
- Correct the OpenSoftware product name in the General privacy copy.

## Root cause

The first refresh implementation reused the Account group's generic in-flow status line for a transient action result. That inserted the message between the group heading and card, shifted the layout, and made a missing `profile:write` grant read like the whole account session had expired. Avatar feedback now uses the global toast surface, and the native bridge maps the legacy-scope rejection to a stable code so the UI can explain that the local change succeeded while network sync did not.

## Validation

- `pnpm typecheck`
- `pnpm exec vitest run src/test/account-avatar.test.ts src/test/app-settings.test.tsx` (96 passed)
- `pnpm test` (2,869 passed, 2 skipped)
- `cargo test --manifest-path src-tauri/Cargo.toml os_accounts::tests` (52 passed)
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets -- -D warnings`
- `pnpm exec biome check` on the touched TypeScript files
- `cargo fmt --manifest-path src-tauri/Cargo.toml -- --check`
- Live Playwright QA at 1180x780: Refresh changed the settings/sidebar geometry together; success and legacy-scope warning rendered as top-right toasts; no Account-group status line or layout shift remained.
- Independent review battery: final Standards, Spec, and Adversarial convergence reviews are clean after checking the #101 contract reconciliation, stale-response/account-lifecycle ordering, snapshot-less upgrades, future-version fallback, and release gating.

- [ ] Needs a June API (backend) deploy before it works end to end. No June API changes are required.

## Deployment dependency

- OS Accounts PR #101 is merged and its API image has deployed successfully to staging. Cross-device refresh in production requires the same API revision there.
- The June OAuth client must allow `profile:write`. Existing cached sessions remain signed in but must sign out and sign in again before a new Avatar selection can sync.
- Staging, RC, stable macOS, and production Windows release workflows now fail closed unless the corresponding `OS_ACCOUNTS_PROFILE_WRITE_READY` attestation is set after a fresh-login smoke test.
- Until both are available, Refresh retains the explicit selection locally and shows a warning toast instead of a false global-success message.

## Out of scope

- User-uploaded profile images.
- Exact palette equality across Apps. Avatar v1 standardizes geometry and lets each host supply semantic theme colors.
- Silently replacing absent or unsupported future seed versions. June derives `v1:default:<User.id>` without a profile write.
- Changes to the app-wide theme or accent system.

## Followups

- Confirm the OS Accounts #101 production deploy, run the fresh-login/Avatar smoke test, and set the production `profile:write` release attestation. Existing sessions may still need one sign-out/sign-in cycle to receive the new scope.

- [x] Security-sensitive changes were called out in the summary. `profile:write` is limited to the User's own `/me` profile; tokens remain in the native keychain and never enter the webview.
- [x] Workflow action references are pinned to full-length commit SHAs. No action references changed.
- [ ] Desktop signing, updater, entitlement, capability, or release changes have owner review. The release workflows add a fail-closed OS Accounts scope attestation and need owner review.
- [x] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. No June API changes are included.


andrewwashuta:
@greptileai I've pushed the review fix.

**Addressed**
- (Greptile) Added an explicit `typeof window === \"undefined\"` guard before avatar localStorage writes, matching the read path.

Validation: Biome, typecheck, and 113 focused tests pass. Please take another look.

andrewwashuta:
@greptileai Please review the latest commit (2247d53c). The new diff adds OS Accounts avatar-seed sync plus the reviewed fallback, logout-lock, and stale-response guards.

andrewwashuta:
@greptileai Please review the latest commit (3a556e8f), including the merged OS Accounts #101 contract reconciliation, toast-based sync feedback, account-lifecycle ordering, and fail-closed profile:write release attestations.

andrewwashuta:
@greptileai Please review the latest commit (f9a8e7cc). A fresh review pass over the branch (8 finder angles, independently verified) produced these changes:

**Fixed**
- **Error mapping:** the exact-message match for a missing scope now runs only after the numeric `error_code` arms (restoring main's precedence), and the mapped copy is self-sufficient ("Sign out and sign in again to update it.") for billing surfaces that render it verbatim and do not branch on `account_permission_required`.
- **Logout lock scope:** the account-operation and refresh locks now cover only the token load+clear; the best-effort revoke POST runs after release, so a slow OS Accounts can no longer pin both global locks for up to 30s during sign-out.
- **Pending-seed reconciliation:** `pendingAccountAvatarAppliesToStoredSeed` no longer treats *any* supported remote seed as compatible, so a newer seed synced from another device wins over a stale failed-sync pending entry, which is now cleared when it stops applying.
- **Settings unmount race:** a sync that resolves after Settings closes now still propagates the updated user to App state (previously the "saved only on this device" hint stayed wrong until the next window-focus refresh). Behavior change: one test asserting the old drop-on-unmount behavior was inverted deliberately.
- **Cleanups:** local variant seed is written only on the localDev path that reads it; the hook uses a single external-store subscription; dead legacy pending-format parse, a needless clone, and a single-use request wrapper removed; `ACCOUNT_AVATAR_CHANGED_EVENT` moved to the `src/lib/events.ts` registry; test isolation now clears both avatar storage prefixes.

**Reviewed and deliberately not changed**
- Status fetch under the account-operation lock stays as-is: it was a deliberate ordering fix from the earlier hardening rounds; a generation-counter redesign is follow-up material, not a pre-merge change.
- The `::before` gradient using the y seed as a horizontal coordinate is intentional variety, and geometry per seed is contract-stable, so it stays.
- Release-gate duplication across the four workflows matches the existing per-workflow required-secrets pattern.

Validation: Biome, typecheck, full frontend suite (2,871 passed), `cargo fmt`/`clippy`/tests all green; local signoff running on f9a8e7cc.

PR #795, opened 2026-07-16
Add Open Software health dashboard and outage monitor

junhohong:
## What changed

- Add a standalone private Next.js health dashboard under `monitor/`.
- Add OS Accounts authorization-code + PKCE login with server-only cookies and an exact server-side user ID allowlist.
- Monitor 16 production surfaces across June, OS API, OS Accounts, Chat services, product portals, and the health dashboard.
- Include non-billable contract probes for dictation cleanup, note generation, and agent chat.
- Add a durable Railway worker that probes every five minutes, retries failures after 15 seconds, and posts only new or changed active outages to Slack.
- Persist outage state on a Railway volume so restarts do not duplicate alerts.
- Add production Dockerfiles and deployment environment documentation.

## Why

Operators need one authenticated production view and a durable outage signal that does not depend on a local Codex automation or generate routine Slack noise.

## Impact

The monitor is isolated under `monitor/`. No existing June application behavior is changed. The authorized user IDs and Slack webhook are deployment secrets and are not committed. Product API probes require the expected unauthenticated contract response, so they do not invoke providers, mutate user data, or consume credits.

The canonical production dashboard is available at `https://health.opensoftware.co`.

## Validation

- `npm test` (13 passing)
- `npm run typecheck`
- `npm run build`
- Production OS Accounts login and allowlist access
- Production dashboard reports 16/16 checks healthy
- Unauthenticated monitor API returns 401
- Railway worker completed with `failures=[]`
- Healthy worker restart produced no Slack message


junhohong:
Fully migrated to the standalone https://github.com/open-software-network/os-health repository. Production now deploys both os-health-web and os-health-worker from os-health@main, and health.opensoftware.co is the only supported dashboard hostname.

PR #798, opened 2026-07-16
Announce RC builds in Slack

junhohong:
## Summary

- send a best-effort Slack webhook after each successful macOS RC publication
- include the RC version, source commit, immutable versioned DMG, and fixed RC release page
- keep fixed updater aliases while making RC provenance lookups fail closed
- document the webhook prerequisite and append the release-channel ADR clarification

## Why

The GitHub Slack subscription announces stable releases because each stable version is newly published. RC builds edit the fixed `rc` prerelease in place, so later candidates do not reliably create Slack updates.

## Validation

- `actionlint .github/workflows/rc-desktop-dmg.yml`
- `git diff --check`
- generated and asserted a sample Slack payload with Node
- verified the GitHub API lookup distinguishes an existing RC release from an explicit 404
- `make local-ci` after every pushed review fix and rebase (frontend and Rust/macOS signoffs correctly reported not applicable)
- standard repo review battery, five adversarial convergence rounds, and final Standards/Spec/Adversarial passes: clean/clean/approve

## Review fixes

- publish and link an append-only versioned DMG so old Slack messages never download a newer RC
- fail closed when the current RC cannot be read, except for an explicit GitHub 404
- upload immutable candidate artifacts without `--clobber`
- bound webhook delivery to one attempt so a lost response cannot create duplicate announcements
- qualify release-channel terminology per `CONTEXT.md`

## Visual testing

Not run. This change only affects release automation and documentation; it has no app UI surface.

## June API deploy

Not required.

## Out of scope

- stable release notifications remain on the existing GitHub Slack app subscription
- the existing multi-asset replacement behavior for fixed RC release channel aliases is unchanged; the Slack step runs only after that publication command succeeds

## Operational note

The production environment must provide `SLACK_WEBHOOK_URL`. Missing or unconfirmed webhook delivery warns without failing an RC whose signed assets already published. Check Slack before manually recovering an unconfirmed delivery.


junhohong:
@codex review

PR #856, opened 2026-07-20
Add one-click morning brief opt-in to onboarding

junhohong:
## What changed
- New final onboarding step (`MorningBriefStep`): a one-click opt-in that creates the existing sandboxed "Morning brief" starter routine (weekdays 8:00 AM) through the normal `createRoutine` path. "Not now" is a first-class quiet path.
- The dictation practice step's button becomes "Continue" since it is no longer the last screen.
- Onboarding preview stub gains the bridge/cron commands so the step can be exercised in a plain browser.
- New CSS for the routine preview card, using existing tokens and the intent-grid card idiom.

## Why
1W retention work: all three June pillars are pull-based, so a fresh install ends with nothing scheduled and no reason to return. The morning brief is the only starter routine whose prompt is fully self-contained (no [placeholder]), which makes a one-click opt-in honest: what you enable is exactly what will run. It manufactures a genuine daily touchpoint on day 1.

## Validation
- `tsc --noEmit` clean
- `pnpm test`: 191 files, 3058 passed
- `src/test/onboarding.test.tsx` extended: full-flow walk now exercises the new step (createRoutine payload asserted), plus decline path and creation-failure path (user stays on the step, error copy shows, quiet path still works)
- Live walkthrough (web preview via in-app browser): step renders at 6/6 with correct copy and schedule ("Weekdays at 8:00 AM"), "Enable morning brief" completes the wizard, "Not now" completes without creating a routine. Screenshot below.

## Known gaps
- Live creation against a real Hermes gateway not exercised in the preview (stubbed); covered by the existing createRoutine plumbing used by Routines templates.

---

### DIFF
--- a/onboarding-preview.html
+++ b/onboarding-preview.html
@@ -351,6 +351,24 @@
               }
               case "p3a_record":
                 return null;
+              // Routine machinery behind the morning brief step. The stub
+              // pretends the bridge is up and creation succeeds, so the
+              // final screen can be exercised end to end in the preview.
+              case "hermes_bridge_status":
+                return { running: true };
+              case "start_hermes_bridge":
+                return { running: true };
+              case "ensure_hermes_bridge_gateway":
+                return null;
+              case "create_hermes_bridge_cron_job":
+                return {
+                  id: "job-morning-brief",
+                  name: args.request && args.request.name,
+                  prompt: args.request && args.request.prompt,
+                  schedule_display: args.request && args.request.schedule,
+                  enabled: true,
+                  state: "scheduled",
+                };
               default:
                 return null;
             }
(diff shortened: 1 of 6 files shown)


junhohong:
QA evidence (web preview walkthrough): the new step at 6/6 with the routine preview card. Enable path completed the wizard and sent the expected createRoutine payload; Not now completed without creating anything.

junhohong:
Addressed all three: the routine name now uses var(--fw-medium) per spec/font-weights.md; enable() checks listRoutines() for an existing "Morning brief" before creating (replay dedupe, covered by a new test); MorningBriefStep routes through goNext so a future step appended after it keeps working.

@codex review

junhohong:
Addressed the P2: when a replay finds an existing Morning brief that is paused or completed, Enable now resumes it through resumeRoutine instead of silently leaving it dormant. Covered by a new test.

@codex review

junhohong:
Addressed the P2: enable() now looks at all Morning brief matches and no-ops when any copy is already scheduled, resuming a dormant one only when nothing is active. New test covers the paused-duplicate-next-to-scheduled case.

@codex review

PR #857, opened 2026-07-20
Notify when a scheduled routine run finishes

junhohong:
## What changed
- New `src/lib/routine-run-notifications.ts`: a pure watch-state module over the scheduled-run session feed. It baselines existing history silently on first sight, then flags each newly ended cron run once. State persists in localStorage so reloads and restarts never renotify; the set is pruned to the fetch window and capped.
- App wiring: a 15s poll (same cadence as the Routines view run history) over `listScheduledRunSessions`, posting one native notification per finished run through the existing `send_app_notification` path, so a click deep-links into the run's conversation via the established `june:agent:open` handshake.

## Why
1W retention work: routines fire on the launchd-managed gateway with zero webview involvement, so today a scheduled routine (like a morning brief) completes in total silence. A routine nobody hears about retains nobody. This closes the loop: scheduled work runs, the user gets told, the click lands them in the result.

## Design notes
- Never backfill: the first poll of an install baselines all already ended runs.
- Freshness gate: an ended run older than 30 minutes at first sight is history, not news (covers overnight run pileups while the app was closed).
- Dedupe survives restarts via persisted state, unlike the in-memory agent-notification dedupe.

## Validation
- `tsc --noEmit` clean
- `pnpm test`: 192 files, 3065 passed
- New `src/test/routine-run-notifications.test.ts` (9 tests): baseline, notify-once, still-running quiet, stale-run quiet, non-cron ignored, fallback copy, pruning, persistence round-trip, corrupted-storage recovery
- Live walkthrough: not run; the trigger requires a real cron run finishing on the gateway. The notification posting and click-through path is the existing `send_app_notification`/`agent_open_ready` machinery already covered by its own tests and prior PR QA.

## Known gaps
- End-to-end proof against a live gateway cron run is manual-pending.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/857"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787166308&installation_model_id=425925&pr_number=857&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F857&signature=852c31d181e2dcd3e0f7c9bcd2061bab51a5df3887d393d1b2c8f40067e5290c"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

### DIFF
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -2069,7 +2069,7 @@ export function App() {
     let state = loadRoutineRunWatchState();
 
     async function poll() {
-      let sessions;
+      let sessions: HermesSessionInfo[];
       try {
         sessions = await listScheduledRunSessions({ includeActive: true });
       } catch {
--- a/src/app/App.tsx
+++ b/src/app/App.tsx
@@ -113,6 +113,7 @@ import {
   agentHudHide,
   agentHudShow,
   agentOpenReady,
+  sendAppNotification,
   type LiveTranscriptEventDto,
 } from "../lib/tauri";
 import { playRecordingSound, preloadRecordingSounds } from "../lib/recording-sounds";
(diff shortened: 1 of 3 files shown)


junhohong:
Addressed both: hasEnded() now requires message_count > 0 so a stuck zero-message cron session never reads as finished (new test); run ids are marked seen only after sendAppNotification resolves via a new markRunsNotified fold, so failed deliveries retry on the next tick (new test).

@codex review

PR #894, opened 2026-07-22
Give the share passcode a one-time field, labels, and a reset path

andrewwashuta:
## Summary

- After creating a passcode-protected share link, the passcode gets a first-class field directly under the link field: masked by default with the same reveal toggle as the passcode input, mono value, and its own copy button - replacing the small "Copy passcode" text button.
- A warning-tone notice under the pair states this is the only time the passcode is shown and to send it separately from the link (June never stores the passcode, so it is unrecoverable once the dialog closes).
- Both fields get "Link" / "Passcode" labels (via the existing `DialogField` primitive) so the two identically-styled rows are distinguishable; plain non-passcode shares keep the original single unlabeled field.
- Reopening the dialog on a passcode-protected link now explains the absence instead of silently dropping the button: "The passcode can't be shown again. To set a new one, reset this link." - where "reset this link" is an inline action that opens a reset-specific confirm dialog, stops the old share, and lands on the create form with the passcode toggle pre-armed.
- `CopyLinkField` gains an optional `id` prop for label association; the stop-sharing confirm state generalizes to `confirmAction: "stop" | "reset"` driving both confirm variants.

## Validation

- `share-dialog.test.tsx`: 13/13 green, including new coverage for the one-time field (masked default, reveal toggle, labels), the reopen caption, and the full reset flow (confirm → `shareDelete` → create form with toggle armed).
- `pnpm typecheck` and `pnpm check` clean.

- [x] Tested visually where it matters (both dialog states walked through in the browser during design review; screenshots reviewed in-session).
- [ ] Needs a June API (backend) deploy before it works end to end. (Frontend-only; uses existing share endpoints.)

## Out of scope

- Changing a passcode in place (re-wrapping the stored content key under a new passcode-derived key without stopping the share). The crypto supports it since the content key is stored locally; the reset path covers the need for now.
- A combined "copy link + passcode" action - deliberately rejected because it undermines the "send it separately" guidance the dialog itself gives.

## Followups

- None planned. If in-place passcode change becomes worth it, the reset confirm flow is the natural place to swap it in.

- [ ] Workflow action references are pinned to full-length commit SHAs. (No workflow changes.)
- [ ] Desktop signing, updater, entitlement, capability, or release changes have owner review. (None.)
- [ ] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. (None.)

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/894"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787294069&installation_model_id=425925&pr_number=894&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F894&signature=fde524ef5bbac51ae22d2f0fd18fc13a974a4677d507855fa5c4987b915a3603"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>

CleanShot 2026-07-22 at 00 48 29@2x

### DIFF
--- a/src/components/share/ShareDialog.tsx
+++ b/src/components/share/ShareDialog.tsx
@@ -285,6 +285,7 @@ export function ShareDialog({
         setInviteId(createdInvite.inviteId);
         setLinkMaterialB64(materialB64);
         setPasswordProtected(Boolean(salt));
+        setRevealPasscode(false);
         try {
           await runClipboardCopy(() =>
             copyExistingLink(
--- a/src/components/share/ShareDialog.tsx
+++ b/src/components/share/ShareDialog.tsx
@@ -335,24 +335,29 @@ export function ShareDialog({
   }, [clearCopyFeedback, passcode, runClipboardCopy, showCopyFeedback]);
 
   const handleStopSharing = useCallback(async () => {
-    if (!shareId) return;
-    await shareDelete(shareId).catch((stopError) => {
+    if (!shareId) return false;
+    try {
+      await shareDelete(shareId);
+    } catch (stopError) {
+      // Swallow the failure so the confirm dialog can close and the error
+      // notice becomes visible in the share dialog, where Stop/Reset remain
+      // available for retry.
       setError(describeShareError(stopError));
-      throw stopError;
-    });
+      return false;
+    }
     setShareId(null);
     setInviteId(null);
     setLinkMaterialB64(null);
     setPasswordProtected(false);
     setLegacyShare(false);
     clearCopyFeedback();
+    return true;
   }, [clearCopyFeedback, shareId]);
 
   // Resetting is stop-sharing plus intent: the owner wants a fresh link with a
   // fresh passcode, so land them on the create form with the toggle already on.
   const handleResetLink = useCallback(async () => {
-    await handleStopSharing();
-    setRequirePasscode(true);
+    if (await handleStopSharing()) setRequirePasscode(true);
   }, [handleStopSharing]);
 
   const handleClose = useCallback(() => {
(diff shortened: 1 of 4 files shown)


andrewwashuta:

**Addressed**
- Reset failure hides its error (Greptile P2) → `handleStopSharing` no longer rethrows on `shareDelete` failure; it sets the error and returns false, so the confirm dialog closes and the error notice is visible in the share dialog with Stop/Reset still available for retry. `handleResetLink` only arms the passcode toggle on success. Added a test covering the failed-reset path (confirm closes, alert visible, link intact, create form not shown).

Please take another look.

andrewwashuta:
@greptileai one more small change from the advisory review, please take another look.

**Addressed**
- Reveal state carried into the one-time passcode field (Codex) → the create-success transition now clears the reveal toggle, so the post-create passcode field always starts masked even if the passcode was revealed while typing it.

PR #890, opened 2026-07-22
Fix model palette activation and search

andrewwashuta:
## Summary

- Open the searchable model palette immediately when Model is selected from the slash menu with a pointer or keyboard.
- Focus model search and remove the staged slash query so subsequent typing filters models instead of editing the message.
- Match model queries across punctuation and intervening catalog terms, including `gpt 5.4` and `opus fast`, in both agent and note-chat pickers.
- Match the existing selector with a plain Auto switch above concrete suggestions, searchable through the Private filter.
- Keep the command palette below the app titlebar at compact window sizes.
- Use the canonical June mark, colored by the active appearance accent, where Auto is represented by a provider badge.

## Root cause

Slash-menu selection inserted `/model ` into the composer and required an undocumented second submit, so model-name typing remained in the message editor. Opening the palette synchronously from the command-row mousedown also raced its outside-click listener, which immediately closed the newly mounted palette in a real browser.

Model search compared the raw normalized query as one literal substring. Catalog names and IDs mix spaces, dots, and hyphens, and may include intervening version terms. The agent and note-chat pickers also maintained duplicate matcher implementations.

The command palette applied the direct-model function-calling capability gate to Auto. Auto is a June-owned private routing choice and its production catalog entry has no direct capability flags, so the generic gate removed it from the palette. Auto now has an explicit router path while concrete models retain the capability gate.

The command palette's height cap measured from the viewport origin without reserving the 28px app titlebar, allowing its top edge to render underneath the titlebar. Auto's logo was maintained as a provider-local `OpenSoftwareMark` copy rather than the canonical, theme-aware June brand component.

## Validation

- `pnpm check`
- `pnpm typecheck`
- Pointer and keyboard regression coverage in `src/test/agent-workspace.test.tsx`
- Production-shaped Auto regression with no direct capability flags in `src/test/model-command-palette.test.tsx`
- Model command palette, note-chat model search, and provider-logo focused suites
- Owning model palette and workspace suites: 348 passed, 2 skipped
- Rendered Chromium QA at 777×556: pointer and keyboard activation focused Search models, model-name search worked, Auto remained searchable with Private enabled, the palette top landed at 41.25px, and Auto matched the existing plain selector toggle above Suggested.
- Search recording: https://app.opensoftware.co/api/v1/files/fil_IHL2MUsr4m17/download

- [x] Tested visually where it matters (rendered interaction QA and recording above).
- [ ] Needs a June API (backend) deploy before it works end to end. No backend deploy is needed.

## Out of scope

[...]

PR #895, opened 2026-07-22
Refine the model picker effort styling and roll in the /model command palette

andrewwashuta:
## Summary

- Replaces the effort meter artwork with a bespoke three-bar SVG (bottom-aligned ascending bars, 2.5px wide, 2px gaps, squared caps); active bars light up per level while the full silhouette stays visible.
- Restyles the composer model trigger: the meter leads the pill, the effort word is dropped (tooltip and aria description still name the level), the Auto preference designation loses its chip container and rides as plain muted text, and the resting trigger darkens to a 65% foreground mix so the model name reads a clear tier above the preference.
- Renames the cost routing preference to "Economy" across the picker, Models settings, and transcript notices ("Auto Economy"); Balanced stays visible in the trigger for consistency.
- Carries the meter into the Effort drill-in rows (each level row leads with its own bar state) and sizes drill-in row values to match their labels.
- Rolls in the /model command palette branch (PR #890): selecting /model from the slash row or submitting bare /model opens a search-first palette over the composer with a pinned Auto toggle, Suggested and All models sections, Private filter, and keyboard navigation; /model <query> still resolves a model directly.

## Validation

- Full agent-workspace suite (344 passed, 2 skipped), palette, app-settings, note-chat, suggested-models, thinking-level, and meter suites all pass; `pnpm typecheck` and `pnpm check` clean.
- Branch merges latest main plus the PR #890 branch; both merges auto-merged and were re-verified after each merge.

- [x] Tested visually where it matters (iterated against live rebuilds during the design pass).

## Out of scope

- The runtime session-state issues identified in the earlier review of the effort feature (pre-send config.set error handling, session-creation ref races, draft fallback relabeling) are deliberately untouched here; this PR is presentation and palette behavior.
- The palette intentionally shows the Auto toggle but not the Preference/Effort drill-in rows; those remain in the model button popover.

## Followups

- Supersedes PR #890; it can be closed once this merges.

---
<a href="https://app.blacksmith.sh/open-software-network/codesmith/os-june/pr/895"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with Codesmith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1787298796&installation_model_id=425925&pr_number=895&repository=open-software-network%2Fos-june&return_to=https%3A%2F%2Fgithub.com%2Fopen-software-network%2Fos-june%2Fpull%2F895&signature=13f4b825e4c58fc72ff7eb83f2a45e676eb3e3a027b0823b09875455c5e87d6d"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with Codesmith</picture></a>


andrewwashuta:

**Addressed**
- Auto Off Selects Missing Fallback (P1) → both the command palette and the toolbar picker now use the curated default id only while the catalog is empty or unloaded; a loaded catalog with no eligible concrete model disables turning Auto off instead of submitting an unresolvable id. Covered by new palette tests for both branches.

Also in this push (user-directed): the palette's Preference and Effort controls now render as inline three-chip rows pinned above the list, mirroring the toolbar picker without its drill-in flyouts, and control adjustments keep the palette open.

Please take another look.

andrewwashuta:
@greptileai pushed a follow-up: the palette now synthesizes the Auto entry when the catalog lacks open-software/auto (the toggle was silently missing on live catalogs), and the Private row's border override was reordered below the base filter rule it was losing to. New palette test covers the synthesized-Auto path.

Please take another look.

andrewwashuta:

**Addressed**
- Auto-only catalog uses fallback → partially, via consolidation: the separate command palette is deleted outright (bare /model now opens the standard composer popover with a root search across both layers), so the duplicated fallback condition is gone. The popover keeps the guard from the prior round: a catalog with concrete entries but none eligible disables Auto-off.

**Skipped (with reason)**
- The explicit catalog-loading state → the host always injects a synthetic entry for the current selection into `options`, so no options-only heuristic can distinguish an unloaded catalog from a loaded Auto-only one; threading a loading flag through every picker surface adds real complexity for a catalog shape the provider never produces (a live catalog always contains concrete models). The remaining fallback only triggers when no concrete entry exists at all, which in practice is the unloaded state, and it keeps the switch usable there. Documented at the fallback site.

Please take another look.

andrewwashuta:
@greptileai one more follow-up pushed: the popover's root search now owns its query independently of the All models flyout's field (typing in the flyout no longer flips the root layer into results mode), with a regression test. Please take another look.

andrewwashuta:
@greptileai addressed the Auto-off fallback finding directly this time: the popover now accepts an explicit `catalogLoaded` signal, and the composer passes its raw catalog state (before the synthetic current-selection entry is injected). The curated-default fallback now applies only while the catalog is genuinely unloaded; a loaded Auto-only catalog disables the switch. Please take another look.

PR #918, opened 2026-07-23
perf(agent): consolidate Hermes lifecycle polling

jakubswierczek:
## Summary

- Consolidates Hermes lifecycle observation into one shared 500 ms `session.active_list` snapshot cycle per active runtime mode, distributing each snapshot to workspace reconciliation and run settlement. Runtime modes schedule independently so a stalled mode cannot throttle a healthy one.
- Uses native persisted history only on bounded degraded paths: after three consecutive unreachable snapshots, after the existing missing-row tolerance, or after a stream disconnect. The current bridge has no revision/delta contract, so this rare fallback remains a full fetch.
- Makes browser approval events primary. Listener registration now retries with capped exponential backoff and surfaces a diagnostic after repeated failures. Initial, listener-reattach, focus, visibility, reconnect, and a 30-second safety snapshot while agent work is active provide recovery without restoring the old permanent 5-second idle poll.
- Adds deterministic coverage for an OPEN socket whose `active_list` frames stop without a close event, native run settlement during that stall, independent runtime-mode cadence, disconnect recovery, listener-rejection approval recovery, approval fallback, and shared-cycle request counts.

## Root cause

Hermes lifecycle consumers owned independent timers: every working workspace refetched its full persisted history every 2.5 seconds, each monitored run polled `session.active_list` every 500 ms, and browser approvals retained a permanent 5-second poll despite an event subscription. Those loops multiplied gateway and persistence work.

The first consolidation also treated an unreachable snapshot as a reason to return from workspace reconciliation and reset run-settlement idle observations. A WebSocket can remain OPEN while frames stop after sleep or a network transition, so a run that completed during that silent stall could remain Working forever. Its single scheduler also awaited both runtime modes together, coupling the healthy mode to the stalled mode's request timeout. Approval recovery was initially one-shot and event-dependent: a rejected listener registration or missed event could hide a pending approval indefinitely in an idle, focused window.

## Validation

- `pnpm check` - passes with the existing ratcheted warning backlog only.
- `pnpm typecheck` - passes.
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test` - all 211 test files and 3,442 tests passed, 2 skipped.
- Focused browser-approval tests - event-first recovery and listener-rejection safety-net recovery pass.
- Focused AgentWorkspace lifecycle matrix - stream disconnect, silent OPEN-socket stall, disappeared run failure, persisted assistant recovery, and active-run preservation all pass.
- `make local-ci` - passes on `29b57bf1cd5c5e55ea44245717370e10fdf33125`: typecheck, all 211 test files and 3,442 tests passed (2 skipped), followed by fresh `signoff/frontend` and not-applicable `signoff/rust-macos` contexts.
- GitHub Actions: repository hygiene and desktop workflows pass on `29b57bf1cd5c5e55ea44245717370e10fdf33125`.
- Greptile re-review: 5/5 on `29b57bf1cd5c5e55ea44245717370e10fdf33125`.
- Not tested visually: this changes lifecycle scheduling and fallback behavior without changing rendered UI or copy; deterministic tests cover the behavior.
- [ ] Tested visually where it matters. Evidence attached below.
- [ ] Needs a June API deploy before this works end to end.

## Out of scope

- Adding a gateway heartbeat. Gateway message deltas are not a lifecycle heartbeat, so silent OPEN-socket stalls use the bounded native fallback in this PR; JUN-414 tracks heartbeat-driven stall detection and reconnect.
- Adding a message revision or delta contract to the Hermes bridge. Until that contract exists, the rare degraded recovery path intentionally performs a full history fetch.
- Changing user-visible run settlement or browser approval behavior.

## Followups

- JUN-414: add a real gateway heartbeat so silently stalled OPEN sockets force reconnect.
- Add a bridge message revision/delta contract before replacing the degraded-path full fetch.

- [x] No secrets, tokens, credentials, or private endpoints are included.
- [x] Security-sensitive changes are called out in **Summary** and have explicit owner review. (No security-sensitive changes.)
- [x] Workflow changes use immutable action references and least-privilege permissions. (No workflow changes.)
- [x] Release notes and user-facing copy avoid unnecessary third-party product comparisons.
- [x] CI is green, or any intentional exception is documented above.

Refs JUN-394

### DIFF
--- a/docs/hermes-architecture.md
+++ b/docs/hermes-architecture.md
@@ -96,6 +96,18 @@ classified events into `AgentChatTurn` / `AgentChatPart[]` for rendering.
   works from classified `JuneHermesEvent`s, never from raw frames. Every
   normalized event carries `receivedAt`; first-party local kinds (`steering`)
   are minted by June and never come out of `classifyHermesEvent`.
+- **Lifecycle reconciliation uses one shared snapshot cycle.**
+  `hermes-active-session-snapshots.ts` requests `session.active_list` once per
+  active runtime mode every cycle and distributes the same result to run
+  settlement and the mounted workspace. The gateway event stream remains the
+  primary message path. Complete persisted history is fetched for a working
+  session only after consecutive missing lifecycle heartbeats or an unexpected
+  stream disconnect; the current bridge has no message-revision or delta
+  contract.
+- **Browser approvals are event-led.** The browser-approval change event
+  refreshes pending approvals promptly. Snapshot reads are limited to initial
+  subscription, listener reattachment, and focus, visibility, or online
+  recovery rather than a permanent timer.
 - **Identity override.** June rewrites the runtime's persona at prompt-build time
   via an injected `SOUL.md`; June presents as June, never as Hermes.
 
--- a/docs/hermes-architecture.md
+++ b/docs/hermes-architecture.md
@@ -96,14 +96,17 @@ classified events into `AgentChatTurn` / `AgentChatPart[]` for rendering.
   works from classified `JuneHermesEvent`s, never from raw frames. Every
   normalized event carries `receivedAt`; first-party local kinds (`steering`)
   are minted by June and never come out of `classifyHermesEvent`.
-- **Lifecycle reconciliation uses one shared snapshot cycle.**
+- **Lifecycle reconciliation uses one shared polling cycle.**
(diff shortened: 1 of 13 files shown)


jakubswierczek:
@greptile review

jakubswierczek:
@greptile review

jakubswierczek:
@greptile review

PR #915, opened 2026-07-23
Fix update relaunch shutdown freeze

jakubswierczek:
## Summary

- Route ordinary quit and in-app update relaunch through one first-request-wins shutdown coordinator.
- Intercept `RunEvent::ExitRequested`, prevent ordinary exit, perform cleanup off the main event loop, and schedule the final exit or restart on the main thread after cleanup or a ten-second aggregate deadline.
- Cover lifecycle paths that bypass `ExitRequested`: the Windows updater runs the same bounded cleanup from `on_before_exit`, while macOS logout/application termination gets a short synchronous `RunEvent::Exit` fallback only when cleanup is still idle.
- Start dictation, Computer use, and Hermes teardown concurrently so the ordered Hermes path always retains its 2 s quiesce + 6 s Gateway + 2 s finalization budget.
- Keep Hermes teardown ordered as quiesce, Gateway unload, runtime reap, then provider proxy shutdown while bounding mutexes, child reaping, external commands, the dictation pipe, and Computer use window coordination.
- Transfer killed children that outlive the bounded poll to detached reapers, give Hermes 500 ms of SIGTERM grace before process-group SIGKILL, and retain bounded browser teardown retries after lock contention.
- Extend the bounded Gateway work from #896 and the dictation reap from #780 without changing the recovery behavior covered by #872.

Refs JUN-410

## Root cause

Update relaunch had competing teardown paths. The updater command synchronously completed its own teardown before requesting restart, and Tauri then entered `RunEvent::Exit`, where June synchronously repeated dictation, Computer use, and Hermes teardown on the main event loop. Several leaves could wait forever: `Child::wait`, mutex acquisition, external `open`/`kill`/version commands, dictation stdin writes, and a synchronous main-thread response wait. Once any duplicate leaf stalled, the relaunch UI was already disabled and the main thread could no longer service macOS, producing the permanent beachball.

Starting cleanup from `RunEvent::Exit` was also too late for a fire-and-forget alternative because Tauri was already tearing down the runtime. Moving the normal path to preventable `ExitRequested` fixed that ordering, but two terminal paths bypass it: `tauri-plugin-updater` calls its Windows `on_before_exit` hook and then `std::process::exit(0)`, while Tao maps macOS logout/application termination directly to `RunEvent::Exit`. Without dedicated coordinator entry points those paths performed no June teardown.

[...]

PR #923, opened 2026-07-23
Fix Computer use run lease lifecycle

jakubswierczek:
## Summary

- Keep the Computer use attended-run lease active across successful `message.complete` transcript segments so the following tool dispatch remains authorized.
- Release the exact lease when the pinned Hermes runtime reports `session.info` with `running: false`, on failed message completion, or on listener teardown.
- Add a replay fixture captured from the real pinned runtime and regression coverage for its plain no-tool terminal sequence.
- Make duplicate native lease releases idempotent and add Rust lease-core tests for duplicate release and successor-run safety.
- Stabilize the native settings event test exposed by the full local CI run by wrapping its React state update in `act`.
- Security-sensitive scope: this corrects the lifetime of the existing attended-run authorization only. It does not widen who can start a lease or which native actions are permitted.

## Root cause

The frontend classified every successful Hermes `message.complete` event as the end of the agent run. Hermes emits that event when an assistant transcript segment is complete, before a following tool is dispatched. The session listener therefore called `computer_use_end_run` and removed the attended-run lease before the first `tool.start`, causing Computer use to reject the action as outside a June chat turn.

Verification against the exact pinned Hermes runtime confirmed the correct boundary: successful `message.complete` is segment completion, and the run-level terminal frame is `session.info` with `running: false`. The fix keeps the lease through successful message completion and releases the exact lease on that terminal session frame. The stored/runtime session ID distinction was not the fault: the stable lease token is derived from the stored session ID plus a UUID, while the runtime session ID only filters inbound events. The JUN-411 readiness probe race and the #910/#915 telemetry and shutdown changes are independent of this early release.

## Validation

- Real pinned-runtime trace: exercised Hermes v2026.7.20, CLI 0.19.0, commit `3ef6bbd201263d354fd83ec55b3c306ded2eb72a`, patch set `june-approval-memory-v14`, through the existing dashboard-gateway smoke protocol with a token-free local provider and the prompt `Reply with the single word: ok`.
- Captured sequence: `session.info running:true` -> `message.start` -> `thinking.delta` -> `message.delta "ok"` -> `reasoning.available` -> `message.complete status:"complete"` -> `session.info running:false`. No `lifecycle.complete` or `turn.completed` frame was emitted; a subsequent `session.active_list` reported the session as `idle`.
- Added the sanitized capture as `src/lib/hermes-control-plane/fixtures/plain-prose-turn.json`; replay coverage asserts that `message.complete` is nonterminal and the final `session.info running:false` frame is terminal.
- `pnpm test:hermes-smoke` was attempted with the exact runtime override, but its macOS LaunchAgent preflight stopped before the gateway phases because the throwaway `HERMES_HOME` has no installed gateway service. The direct dashboard-gateway portion of the existing smoke harness completed successfully and produced the trace above.
- Added Rust lease-core unit coverage for `begin_run` / `end_run` / `stop_if_idle`, including duplicate release idempotency and protection against stale cleanup stopping a successor run.
- Red/green frontend regression: the lease survives successful `message.complete` and releases on `session.info running:false`.
- `pnpm check`
- `pnpm typecheck`
- `pnpm test`: 209 files passed, 3,426 tests passed, 2 skipped.
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make tauri-fmt-check tauri-lint tauri-test`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer NODE_OPTIONS=--no-experimental-webstorage CI=true make verify`
- `DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer NODE_OPTIONS=--no-experimental-webstorage CI=true make local-ci`
- `signoff/frontend` and `signoff/rust-macos` recorded on `4d19db640f1bc16e27c7d6a97375f9f893db8332`.
- `prepare-cua-driver.mjs` was not run because the Computer use driver was not changed.

- [ ] Tested visually where it matters (UI changes: attach a screenshot or recording). No rendered UI changed.
- [ ] Needs a June API (backend) deploy before it works end to end. No backend deploy is needed.

## Out of scope

- JUN-411 permission-readiness probe stampede work.
- The lease heartbeat/TTL backstop tracked by JUN-414.
- Computer use driver changes or permission setup.
- Changes to #910 telemetry or #915 shutdown behavior.

## Followups

- JUN-411 remains the separate followup for permission-readiness probe coordination.
- JUN-414 tracks the lease heartbeat/TTL backstop; it is deliberately not implemented here.

- [x] Workflow action references are pinned to full-length commit SHAs. No workflow action references changed.
- [x] Desktop signing, updater, entitlement, capability, or release changes have owner review. Not applicable; none changed.
- [x] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. Not applicable; none changed.
- [x] User-facing copy follows the repo UI conventions. No user-facing copy changed.

Refs JUN-413

### DIFF
--- a/src-tauri/src/computer_use.rs
+++ b/src-tauri/src/computer_use.rs
@@ -1658,6 +1658,10 @@ pub fn computer_use_begin_run(
         ));
     }
     let session_id = validate_run_session_id(&request.session_id)?;
+    begin_run_lease(&state, session_id)
+}
+
+fn begin_run_lease(state: &ComputerUseState, session_id: String) -> Result<(), AppError> {
     let mut runs = state
         .attended_runs
         .lock()
@@ -1677,10 +1681,19 @@ pub fn computer_use_begin_run(
     if inserted {
         state.attended_generation.fetch_add(1, Ordering::SeqCst);
         if starts_new_task {
-            clear_app_authorizations(&state);
+            clear_app_authorizations(state);
         }
     }
     Ok(())
+}
+
+fn end_run_lease(state: &ComputerUseState, session_id: &str) -> Result<Option<u64>, AppError> {
+    let mut runs = state
+        .attended_runs
+        .lock()
+        .map_err(|_| AppError::new("computer_use_unavailable", "Run lease lock failed."))?;
+    let removed = runs.remove(session_id);
+    Ok((removed && runs.is_empty()).then(|| state.attended_generation.load(Ordering::SeqCst)))
 }
 
 #[tauri::command]
@@ -1690,14 +1703,7 @@ pub async fn computer_use_end_run(
     request: ComputerUseRunRequest,
 ) -> Result<(), AppError> {
     let session_id = validate_run_session_id(&request.session_id)?;
-    let idle_generation = {
-        let mut runs = state
-            .attended_runs
-            .lock()
-            .map_err(|_| AppError::new("computer_use_unavailable", "Run lease lock failed."))?;
-        let removed = runs.remove(&session_id);
-        (removed && runs.is_empty()).then(|| state.attended_generation.load(Ordering::SeqCst))
-    };
+    let idle_generation = end_run_lease(&state, &session_id)?;
     if let Some(generation) = idle_generation {
         stop_if_idle(&app, &state, generation).await;
     }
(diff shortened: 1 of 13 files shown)


jakubswierczek:
@greptile review

PR #917, opened 2026-07-23
perf(notes): debounce persistence and patch updates

jakubswierczek:
## Summary

- Debounce note title and content persistence per note while keeping the editor optimistic.
- Flush pending note saves on blur, note navigation, deletion, relaunch, and coordinated app shutdown.
- Add an opt-in patch response to `update_note` and implement the hot path as one `UPDATE ... RETURNING` SQL statement while preserving the legacy response for existing callers.
- Integrate pending note saves into the shared shutdown coordinator introduced by PR #915.

Refs JUN-389

## Root cause

The note editor persisted ordinary title and content changes immediately. Each `update_note` call also hydrated the current note, wrote it, and hydrated it again, so routine typing created avoidable command and database traffic. Delaying writes without a shutdown barrier would have introduced a data-loss window, so the save queue now participates in the coordinated shutdown handshake.

## Validation

- `pnpm check`
- `pnpm typecheck`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test` (210 files passed; 3,427 tests passed; 2 skipped)
- `cargo fmt --manifest-path src-tauri/Cargo.toml --all -- --check`
- `cargo clippy --manifest-path src-tauri/Cargo.toml --all-targets --locked -- -D warnings`
- `cargo test --manifest-path src-tauri/Cargo.toml --locked`
- `CI=true make local-ci` (`signoff/frontend` and `signoff/rust-macos` succeeded on `97deb44a74b72968a9cb6082c2d470380e5e2a03`)

- [ ] Tested visually where it matters. No visible UI behavior or styling changed; automated editor and shutdown tests cover the save semantics.
- [ ] Needs a June API (backend) deploy before it works end to end. No backend deploy is required.

## Out of scope

- Removing or changing the legacy `update_note` response contract.
- Broader changes to note generation, transcription, or storage schemas.
- Changing the debounce interval into a user-configurable setting.

## Followups

- Keep this PR in draft for the requested Opus + Kimi review round before marking it ready.

## Security and release checklist

- [x] No secrets, local `.env` values, signing material, tokens, or customer data are committed.
- [x] Security-sensitive changes were called out in the summary. No security-sensitive behavior changed.
- [x] Workflow action references are pinned to full-length commit SHAs. No workflow action references changed.
- [ ] Desktop signing, updater, entitlement, capability, or release changes have owner review. Not applicable.
- [ ] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. Not applicable.
[...]

PR #925, opened 2026-07-23
fix(notes): preserve concurrent edits after generation

jakubswierczek:
## Summary

- Preserve user-owned note title and edited content when generation completes from a stale note snapshot.
- Keep the existing generation behavior that appends into unchanged, non-null edited content.
- Add a deterministic SQLite-trigger regression test that persists a user edit between generation's read and final update.
- Add a second trigger regression for a non-null snapshot that diverges before the final update.
- Refs JUN-412.

## Root cause

`set_generated_note_for_session` read the note, derived a complete replacement row, and later wrote the stale `edited_content` value back. If autosave persisted an edit after the read, the generation-complete update could overwrite it, including replacing the concurrent edit with `NULL`. The same stale-row update could also replace a concurrent title edit.

The final update now uses compare-and-set guards for the user-owned columns. A `NULL` edited-content snapshot never authorizes generation to write that column; an existing edited body is updated only if it still exactly matches the snapshot that generation read. The title follows the same snapshot guard.

## Validation

- `env CI=true NODE_OPTIONS=--no-experimental-webstorage DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make verify`
- `env DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make tauri-fmt-check tauri-lint tauri-test` after merging current `origin/main` at `3e3890d5f45cfd59c285f3d0a3d1ac98ffca16c7`
- `env CI=true NODE_OPTIONS=--no-experimental-webstorage DEVELOPER_DIR=/Applications/Xcode.app/Contents/Developer make local-ci`
- Regression test: `generated_note_preserves_user_edit_written_after_its_read`
- Regression test: `generated_note_preserves_newer_non_null_user_edit_written_after_its_read`
- `signoff/frontend` posted as not applicable by path.
- `signoff/rust-macos` passed on `2aae53b605d39633de29c7d378e57c8cb42fcecd`.
- The first `local-ci` attempt hit a transient `WouldBlock`/connection-reset failure in the unrelated loopback redirect test. That test passed immediately in isolation, and the complete `local-ci` rerun passed.

- [x] Tested visually where it matters (not applicable: this is a database concurrency fix with no UI changes).
- [ ] Needs a June API (backend) deploy before it works end to end. No June API deploy is needed.

## Out of scope

- Changing the autosave behavior in #917.
- Changing generated-content composition or session-block replay semantics.
- Adding a schema-level row version; exact-value compare-and-set guards are sufficient for the owned columns in this path.

## Followups

- Run the requested dual review round while this PR remains a draft.
- #917 landed in `origin/main` before this final merge. This fix remains compatible with its continuous-autosave path.
- The reverse-direction autosave-clobbers-generated-title race is tracked in JUN-418; do not fix it here.
- The SQLite busy-timeout durability gap is tracked separately in JUN-407; do not fix it here.

- [x] Security-sensitive changes were called out in the summary (not applicable).
- [x] Workflow action references are pinned to full-length commit SHAs (no workflow changes).
- [x] Desktop signing, updater, entitlement, capability, or release changes have owner review (not applicable).
- [x] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review (not applicable).
- [x] User-facing copy follows the repo UI conventions (no user-facing copy changes).

### DIFF
--- a/src-tauri/src/db/repositories.rs
+++ b/src-tauri/src/db/repositories.rs
@@ -3119,6 +3119,7 @@ impl Repositories {
         } else {
             append_note_content(current.generated_content.clone(), content.clone())
         };
+        let expected_edited_content = current.edited_content.clone();
         let next_edited_content = current.edited_content.map(|edited_content| {
             if existing_session_block {
                 if edited_content.trim()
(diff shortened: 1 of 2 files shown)


jakubswierczek:
@greptile review

jakubswierczek:
@greptile review

PR #940, opened 2026-07-23
refactor(db): version SQLite migrations

jakubswierczek:
## Summary

- Replaces launch-time full-schema replay with an append-only 29-version `schema_migrations` catalog ordered by when each change reached `main`.
- Adopts existing unversioned installs by introspecting tables, indexes, and columns, stamping only the contiguous schema prefix that is actually present, and then applying pending versions.
- Runs pending schema statements and version writes together in one `BEGIN IMMEDIATE` transaction, with grouped one-inspection-per-table additive column repairs.
- Keeps every existing SQL migration file untouched and records the versioning and adoption contract in ADR 0037.
- Security and data-safety note: this changes the critical startup path for every local database and deliberately fails closed on unknown or gapped legacy schemas.

## Root cause

`run_migrations` had no durable record of completed schema work. It replayed nearly every historical SQL file on each process launch and guarded additive columns with one `PRAGMA table_info` query per column. The only protection around the two destructive deduplication migrations was checking for their indexes on every launch.

A simple "create a version table and stamp latest" conversion would be unsafe: released databases have no ledger, and a genuinely older database would be mislabeled current. Starting all unversioned databases at zero would be unsafe in the opposite direction because it would rerun the transcript and agent-message deduplication statements. The historical SQL filenames also cannot be treated as a trustworthy sequence because parallel branches produced duplicate `014_*` names and later renumbering.

This PR reconstructs the authoritative order from first-parent `main` history and uses explicit schema landmarks to derive the exact contiguous prefix before stamping. A fresh database has no known prefix, so it runs the whole catalog. An unknown or noncontiguous schema fails closed instead of guessing.

## Validation

- `make verify` passed:
  - Biome check and TypeScript typecheck.
  - Vitest: 216 files, 3,504 passed, 2 skipped, 0 failed.
  - Tauri `cargo fmt`, `cargo clippy --all-targets --locked -- -D warnings`, and the full Cargo test suite.
  - June API formatting, Clippy, and full test suite.
- `make local-ci` passed on `d0a539906a395cb18ed402d0932e03d3b859da43`:
  - `signoff/frontend` posted as not applicable because no frontend-signoff paths changed.
  - `signoff/rust-macos` posted successfully after formatting, Clippy, and the full macOS Rust suite.
- Migration-specific Rust tests cover:
  - A fresh database running all 29 versions and stamping latest.
  - A current replay-built database with no ledger stamping latest without replaying destructive SQL; a guard trigger would abort if destructive replay occurred.
  - An unversioned migration-15 database upgrading cleanly.
  - Every historical unversioned schema prefix upgrading cleanly.
  - A deliberately failing pending migration rolling back all pending schema and version changes while preserving the previously committed version.
  - Catalog validation rejecting migrations without legacy schema requirements.
- The existing concurrent-startup integration test passes with multiple migration runners.
- Existing SQL migration files are unchanged.

- [x] Tested visually where it matters. Not applicable: this is a low-level SQLite startup refactor with no user-visible UI surface; validation is at the SQLite and Rust boundary.
- [ ] Needs a June API (backend) deploy before it works end to end. No June API deploy is required.

## Out of scope

- Startup recovery scan and read batching, plus per-recovery write batching.
- Consolidating the duplicated agent-task repair call.
- Changing or renumbering any existing SQL migration file.

## Followups

- Consolidate startup recovery reads and writes and the duplicated task repair in a separate change; those are behaviorally independent from the migration-version adoption boundary.

Refs JUN-393

- [x] Workflow action references are pinned to full-length commit SHAs. No workflow changes.
- [ ] Desktop signing, updater, entitlement, capability, or release changes have owner review. Not applicable.
- [ ] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review. Not applicable.
- [x] User-facing copy follows the repo UI conventions. No user-facing copy.

### DIFF
--- a/docs/adr/0037-versioned-local-sqlite-migrations.md
+++ b/docs/adr/0037-versioned-local-sqlite-migrations.md
@@ -1,0 +1,78 @@
+# ADR 0037: Version June's local SQLite schema with a release-ordered catalog
+
+Date: 2026-07-23
+Status: accepted
+
+## Context
+
+June historically built its local SQLite schema by replaying almost every
+migration statement at every process launch. Additive columns were repaired by
+individual `PRAGMA table_info` probes, one probe per column. This made launch
+cost grow with the entire schema history and kept destructive deduplication
+statements on the startup path.
+
+Existing installations have no migration ledger. They may contain the complete
+schema produced by the replay runner or an older, contiguous historical schema.
+Blindly assigning those databases the latest version could skip a real pending
+migration. Replaying from version zero could rerun data-changing statements.
+
+The SQL filenames cannot be used as the version authority. Parallel feature
+branches produced two distinct `014_*` files, and several files were renumbered
+after their schema changes had already reached main.
+
+## Decision
+
+June owns a `schema_migrations` table with one row per applied version. A
+release-ordered catalog in `src-tauri/src/db/migrations.rs` is the authoritative
+sequence. Its integer versions are contiguous and append-only; the descriptive
+name records intent without treating historical filename prefixes as versions.
+
+On a versioned database, June reads the ledger and returns immediately when it
+already matches the catalog. When migrations are pending, June applies every
+pending catalog entry and records its row inside one `BEGIN IMMEDIATE`
+transaction. A failed statement rolls back both schema changes and version rows.
+
+On the first versioned launch of an unversioned database, June takes one schema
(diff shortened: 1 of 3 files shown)


jakubswierczek:
@greptile review

PR #943, opened 2026-07-23
Recover idle Gateway stalls on submit

jakubswierczek:
## Summary

- Before the first mutating RPC of an idle submit, send one read-only `session.active_list` liveness request with a three-second deadline.
- If that safe preflight times out, force-disconnect the target runtime mode, reconnect, and retry only the preflight once.
- Send `session.create`, `session.resume`, and `prompt.submit` at most once with their normal deadlines; concurrent same-mode idle submits share the preflight.
- Keep the healthy and active-submit paths unchanged without adding idle polling or changing heartbeat counters.

## Root cause

The shared `session.active_list` heartbeat only runs while working-session consumers are subscribed. A silently stalled OPEN socket could therefore remain undetected while idle, and the next submit inherited the 120-second default timeout. The original recovery wrapped whichever RPC happened to be first, which could be non-idempotent `prompt.submit` for a cached runtime session or `session.create` for a new one. Retrying either after a false three-second timeout could duplicate a run, credits, or an empty session.

## Validation

- `pnpm check`
- `pnpm typecheck`
- `NODE_OPTIONS=--no-experimental-webstorage pnpm test` (216 files; 3,514 passed, 2 skipped)
- `make local-ci` on `018069c5de1bdd5da55d20267f8a3a55b693e528` (`signoff/frontend` passed; `signoff/rust-macos` passed as not applicable)
- Fault-injection coverage verifies cached-runtime idle-stall recovery in under 10 seconds with exactly one `prompt.submit`, a live four-second submit acknowledgement without reconnect or double-submit, new-session recovery with exactly one `session.create`, a surfaced second preflight failure, and shared recovery for concurrent same-mode submits.

- [ ] Tested visually where it matters (not applicable: transport-only behavior with no UI changes).
- [ ] Needs a June API (backend) deploy before it works end to end (no: desktop-only change).

## Out of scope

- Always-on idle polling or any new periodic Gateway traffic.
- Changes to heartbeat miss counters or working-session settlement.
- June API changes.

## Followups

None.

- [x] Security-sensitive changes were called out in the summary (not applicable: no security boundary changed).
- [x] Workflow action references are pinned to full-length commit SHAs (not applicable: no workflow changes).
- [x] Desktop signing, updater, entitlement, capability, or release changes have owner review (not applicable).
- [x] Backend auth, billing, metering, CORS, rate limit, or logging changes have owner review (not applicable).
- [x] User-facing copy follows the repo UI conventions (not applicable: no UI copy changes).

Refs JUN-433

### DIFF
--- a/docs/hermes-architecture.md
+++ b/docs/hermes-architecture.md
@@ -125,7 +125,11 @@ classified events into `AgentChatTurn` / `AgentChatPart[]` for rendering.
   heartbeat: three consecutive request timeouts invalidate every open client
   for that runtime mode, converting a silently stalled OPEN socket into the
   established unexpected-close and reconnect path without adding another
-  periodic request.
+  periodic request. When that polling is dormant because the mode has no
+  working sessions, the first Gateway request of a new submit instead gets a
+  three-second liveness deadline. A timeout force-disconnects the mode and
+  retries that transport request once on the fresh connection; the rest of the
+  submit keeps ordinary request deadlines.
 - **Browser approvals are event-led.** The browser-approval change event
   refreshes pending approvals promptly. Snapshot reads are limited to initial
   subscription, listener reattachment, and focus, visibility, or online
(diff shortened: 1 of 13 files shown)


jakubswierczek:
@greptile review

PR #1016, opened 2026-07-31
fix(agent): surface actionable run failures

junhohong:
## Summary

- classify agent failures as tool, model provider, local runtime, context, or credits failures before persistence
- show a specific user-facing notice instead of the generic "June could not complete this request"
- restore the direct in-app issue report dialog and offer a removable diagnostic attachment with only the latest failed run's bounded IDs, stable error code, and completion time
- fence crash settlement to the sidecar instance that owns each run, preserve credit failures, and keep ambiguous mutating tool failures non-retryable

## Root cause

The TypeScript agent runtime flattened every uncancelled exception to a string. The Rust host then persisted every failure as `agent_run_failed` with `retryable: true`, and the frontend interpreted every retryable error as an upstream provider outage. This discarded whether a tool, provider, context limit, credit gate, or local runtime had actually failed. The workspace rebuild also stopped consuming the existing report-category navigation marker, so the direct report entry points no longer opened the in-app dialog.

Review also exposed two lifecycle bugs in the initial fix: tool exceptions were converted to non-fatal model-visible results by the SDK, and runtime EOF used synthetic identities that the selected workspace ignored. The final implementation makes host execution failures terminal while leaving malformed model arguments self-correctable, and settles only runs owned by the exact sidecar instance that stopped. Transport-ambiguous dispatches retain ownership; a timed-out owner is UUID-fenced and terminated so its runs receive a real terminal event.

## Verification

- `pnpm agent-runtime:typecheck`
- `pnpm agent-runtime:test` (62 passed)
- targeted frontend and native ownership/privacy regressions
- full frontend suite (154 files, 1,669 tests passed)
- `cargo check --manifest-path src-tauri/Cargo.toml`
- Rust desktop fmt, clippy, and full tests
- companion fmt, clippy, and tests
- June API fmt, clippy, and tests
- full `make verify`
- clean Standards, Spec, and adversarial review convergence
- live web-preview walkthrough: Account -> Report a bug opened the preselected modal, accepted a description, and enabled Send report

## Visual test

PASS. The restored issue-report modal was exercised in the local June preview. Layout, category selection, focus, description entry, and enabled submission state rendered correctly.

## Backend deploy needed

No. Diagnostics use the existing issue-report attachment contract, so the stable desktop can ship independently of June API.

## Out of scope

- exposing raw exception strings in the user-facing transcript
- attaching prompts, conversation content, tool output, or free-form persisted error messages to reports

## Follow-ups

- none required for this release


junhohong:
@greptile-apps please review the current head.

junhohong:
@codex review

junhohong:
@codex review

junhohong:
@greptile-apps please review commit 78e51e916a927cd48c8d841c0b2b383a14a4ac18.

